PR #11, opened 2015-03-25
Add json output formatter

shakhat:
The patch extends output parameter with option "json". With it netperf-wrapper prints raw json data to stdout.

### DIFF
--- a/netperf_wrapper/formatters.py
+++ b/netperf_wrapper/formatters.py
@@ -202,6 +202,24 @@ class CsvFormatter(TableFormatter):
         for row in self.combine_results(results):
             writer.writerow(list(map(format_item, row)))
 
+
+class JsonFormatter(Formatter):
+    """Format the output as json."""
+
+    def format(self, results):
+        self.open_output()
+
+        data = []
+        for result in results:
+            d = result.serialise()
+            for s in sorted(result.series_names):
+                units = self.settings.DATA_SETS[s]['units']
+                d['metadata']['SERIES_META'][s]['UNITS'] = units
+            data.append(d)
+
+        self.output.write(json.dumps(data))
+
+
 class StatsFormatter(Formatter):
 
     def __init__(self, settings):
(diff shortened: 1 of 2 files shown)


tohojo:
What's the use case? The data files are already json...

shakhat:
Use case is following: My tool runs netperf-wrapper on remote machine as a single command (e.g. "ssh <host> netperf-wrapper <params>"). The tool captures stdout and then calculates stats on the local machine. 
Currently netperf-wrapper writes json into a separate file which makes it harder to unitilize it as one command (the possible workaround is to run netperf-wrapper without formatter, grep out file name and then cat this name onto stdout)

tohojo:
Right, that makes sense. I'd prefer it if there wasn't a difference in
the output generated by this and the results file, though. Storing the
units in the metadata is probably a good idea in any case, so that
should be added to the resultset somewhere. Will think about where it
makes the most sense.

Also, to make this be compatible with the results format, the outermost
json object shouldn't be wrapped in a list if there's only one
resultset. And detecting a .gz extension and applying gzip compression
would be nice.

Feel free to resubmit a pull request with the above changes. Otherwise,
I'll put it on my list. If you do resubmit, please add a Signed-Off-By:
header to the commit message.

Thanks,
-Toke

shakhat:
Ok, I'll update the patch accordingly. It would be even better to split it into two: one for units in json data file and other for json formatter.

Regards,
-Ilya

tohojo:
> Ok, I'll update the patch accordingly. It would be even better to split it into
> two: one for units in json data file and other for json formatter.

Great! You're right that splitting them up makes sense; but feel free to
do them as two commits in one pull request if that is less of a hassle :)

The 'units in the results' part should probably go in
Aggregator.collect() somewhere...

-Toke

tohojo:
Closing for now. Please open a new pull request if you re-do the patch :)

PR #31, opened 2015-05-10
tcp1_down: Fix copy+pasted description

sourcejedi:
tcp1_down is not bidirectional :).


tohojo:
Quite right. Thanks for noticing :)

PR #40, opened 2015-05-24
Capture notsent_lowat limit_output_bytes and timestamp sysctl settings

dtaht:
tcp_limit_output_bytes is a very critical TSQ parameter. Originally
(and how it is deployed at scale) this is set very low (4k or so)
and used in combination with sch_fq to spectacular results.

Unfortunately this low setting messed with tcp on aggregating wifi device
drivers and in the linux kernel mainline it was eventually bumped
to 128k by default.

This has really negative results on lots of flows at gigE rates with
sch_fq, bloating up to thousands of packets with hundreds of flows
extant.

notsent_lowat is usually turned off universally because it causes more
context switches. it is a good idea to have on in many scenarios, however.
I have really conflicting information as to whether and where it is being used.

timestamps were often turned off on older routers for performance reasons.
various tcps will behave differently with timestamps on or off.

Capturing all this info will help examine these effects.


tohojo:
Merged in commit 73fc278.

PR #54, opened 2015-07-28
Docs should be updated now output files are in /tmp

sourcejedi:
(but I liked the old location :p).  I noticed this one instance, I don't know if there are other mentions.


sourcejedi:
and I think I got the markdown wrong, sorry.  Just consider this an issue quoting where the docs are out of date.

tohojo:
Thanks for noticing! However, the git version reverts back to putting the files in the current directory by default (but with a command line switch to change that).

Will put out a new version soon, which should fix this (thought I had already, actually).

sourcejedi:
yay thx :)

PR #61, opened 2016-02-08
Multiple new test configurations for multiple stream rrul, tcp_down, tcp_up combinations -- 2nd attempt

cklatsky:
New test configurations for multiple stream rrul, tcp_down, and tcp_up combinations.  rrul_X_up_Y_down are based off of rrul_be.  They have X upstream streams, Y downstream streams, and Y ping streams.  tcp_Xdown is based off of tcp_down.  They have X downstream streams and one ping stream.  tcp_Yup is based off of tcp_up.  They have Y upstream streams and one ping stream.


dtaht:
My principal objection to this is that with the addition of so many tests that other users attempting to get comparable data will be confused as to what valid tests to run. Merely getting people to burn 5 minutes on the basic tcp_upload, download, rrul, rrul_be, and rtt_fair tests has been a problem. Certainly I welcome a higher density test (like 12 down). But I would like to have people adopt other tests that test somewhat realistic and different workloads (like rtt_fair, tcp_2up_delay, web, and the voip tests do), than go nuts running all these other bulk oriented tests.  

I've had a few other tests in the queue (tcp_staircase) that I should get around to finishing, and I do keep hoping to get something simpler for voip/gaming traffic simulation than d-itg.

All that said, I've tested a few of these, they work, and I'm really big on having comparable data, so I approve....

tohojo:
Pushed a set of changes that adds tcp_nup and tcp_ndown tests which support specifying the number of streams as a test parameter. I have then added tcp_{2,4,6,8,12}{up,down} tests that simply sets the number of streams for the *n* tests. This reduces code reuse and is hopefully more flexible.

Could you confirm that these tests work for your needs? If so, I'll add a bidirectional (RRUL) variant following the same principle.

cklatsky:
Sure, I will try these out.  I should have time later today to try them.

cklatsky:
I was able to try out both the tcp_nup & tcp_ndown test configurations.  Both work well, thanks for coding that up.  I noticed that in the plots available when reviewing the results are a reduced set of plots.  Any reason not to make the full set of plots available for all test results files?  Might be a good candidate to have them all in a separate file and just include that file for each new test configuration?

tohojo:
Okay, pushed an rrul_be_nflows test that supports specifying the number of flows in each direction. Care to take that for a spin as well? Since there's already a whole bunch of RRUL tests, my inclination would be to not do derived tests for the specified number of flows in this case; can you live with specifying the --test-parameter options yourself?

And yeah, more reuse of plot configurations has been on my list for a while. However, since the plots specify which data sets they use, it's not always possible to reuse them across different tests. But it should be possible achieve a bit more reuse than the current state...

cklatsky:
Yes, its not an issue to specify the --test-parameter.  I will give this updated file a try tomorrow when I get back to the office and let you know.  Thanks.

cklatsky:
I was able to test rrul_be_nflows today, using different combinations of upstream & downstream streams.  Everything worked fine.  Thanks for making this file.  If you are OK with this feedback, I think this pull request can be closed.

tohojo:
Excellent! Thanks for testing!

And yeah, closing this now :)

PR #91, opened 2017-01-31
introduce ss iteration script

cherusk:
more efficient core yet to come

### DIFF
--- a/flent/runners.py
+++ b/flent/runners.py
@@ -682,6 +682,14 @@ class NetperfDemoRunner(ProcessRunner):
             args['socket_timeout'] = ""
 
         if args['test'] in ("TCP_STREAM", "TCP_MAERTS", "omni"):
+            self.ctrl_port = args['control_port']
+            ss_args = {'host': self.remote_host or 'localhost',
+                       'interval': args['interval'],
+                       'length': args['length'],
+                       'target': args['host'],
+                       'ip_version': args['ip_version']}
+
+            self.add_child(SsRunner, ss_args, self.delay, None)
             args['format'] = "-f m"
 
         return "{binary} -P 0 -v 0 -D -{interval:.2f} -{ip_version} {marking} " \
@@ -950,6 +958,137 @@ class IperfCsvRunner(ProcessRunner):
                     "an --enhancedreport option. Not using.")
 
         raise RuntimeError("No suitable Iperf binary found.")
+
+
+class ParseException(Exception):
+    pass
+
+
+class SsRunner(ProcessRunner):
+    """Runner for iterated `ss -t -i -p`. Depends on same partitial output
+    separationa and time stamping as TcRunner."""
+
+    ip_v4_addr_sub_re = "([0-9]{1,3}.){3}[0-9]{1,3}(:\d+)"
+    # ref.: to commented, untinkered version: ISBN 978-0-596-52068-7
+    ip_v6_addr_sub_re = "(?:(?:[A-F0-9]{1,4}:){7}[A-F0-9]{1,4}|\
+                         (?=(?:[A-F0-9]{0,4}:){0,7}[A-F0-9]{0,4})\
+                         (([0-9A-F]{1,4}:){1,7}|:)((:[0-9A-F]{1,4})\
+                         {1,7}|:))(:\d+)"
+
+    time_re = re.compile(r"^Time: (?P<timestamp>\d+\.\d+)", re.MULTILINE)
+    cwnd_re = re.compile(r"cwnd:(?P<cwnd>\d+)", re.MULTILINE)
+    rtt_re = re.compile(r"rtt:(?P<rtt>\d+\.\d+/\d+\.\d+)", re.MULTILINE)
+    pid_re = re.compile(r"pid=(?P<pid>\d+)", re.MULTILINE)
+    ports_ipv4_re = re.compile(r"" + "(?P<src_p>" + ip_v4_addr_sub_re + ")" +
+                               "\s+" + "(?P<dst_p>" + ip_v4_addr_sub_re + ")")
+    ports_ipv6_re = re.compile(r"" + "(?P<src_p>" + ip_v6_addr_sub_re + ")" +
+                               "\s+" + "(?P<dst_p>" + ip_v6_addr_sub_re + ")",
+                               re.IGNORECASE)
+    ss_header_re = re.compile(r"" + "State\s+Recv-Q\s+Send-Q\s+Local")
+
+    src_p = []
+    dst_p = []
+    # ref.: upstream ss
+    ss_states = ["UNKNOWN", "ESTAB", "SYN-SENT", "SYN-RECV", "FIN-WAIT-1",
+                 "FIN-WAIT-2", "TIME-WAIT", "UNCONN", "CLOSE-WAIT", "LAST-ACK",
+                 "LISTEN", "CLOSING"]
+    ss_states_re = re.compile(r"|".join(ss_states))
+
+    def __init__(self, *args, **kwargs):
+        super(SsRunner, self).__init__(*args, **kwargs)
+        self.ctrl_port = str(self._parent.ctrl_port)
+
+    def filter_np_parent(self, part):
+        sub_part = []
+        sub_parts = self.ss_states_re.split(part)
+        sub_parts = [sp for sp in sub_parts if sp.strip()
+                     and not self.ss_header_re.search(sp)]
+
+        for sp in sub_parts:
+            pid = self.pid_re.search(sp)
+            if None is pid:
+                continue
+            pid_str = pid.group('pid')
+
+            f_ports = self.ports_ipv4_re.search(sp)
+            if None is f_ports:
+                f_ports = self.ports_ipv6_re.search(sp)
+
+            if None is f_ports:
+                raise ParseException()
+
+            dst_p = str(f_ports.group('dst_p').split(":")[-1])
+
+            if self.par_pid == pid_str and dst_p != self.ctrl_port:
+                sub_part.append(sp)
+
+        if 1 != len(sub_part):
+            raise ParseException()
+
+        return sub_part[0]
+
+    def parse_part(self, part):
+        sub_part = self.filter_np_parent(part)
+
+        timestamp = self.time_re.search(part)
+
+        cwnd = self.cwnd_re.search(sub_part)
+        rtt = self.rtt_re.search(sub_part)
+
+        if None in [timestamp, cwnd, rtt]:
+            raise ParseException()
+
+        timestamp = float(timestamp.group('timestamp'))
+        cwnd = int(cwnd.group('cwnd'))
+        rtt, rtt_var = [float(x) for x in rtt.group('rtt').split('/')]
+
+        raw_values = {'t': timestamp, 'cwnd': cwnd, 'rtt': rtt, 'rtt_var': rtt_var}
+        self.raw_values.append(raw_values)
+
+        return raw_values
+
+    def parse(self, output):
+        self.par_pid = str(self._parent.pid)
+        results = {}
+        parts = output.split("\n---\n")
+        for part in parts:
+            try:
+                res_dict = self.parse_part(part)
+                t = res_dict['t']
+                for k, v in res_dict.items():
+                    if k == 't':
+                        continue
+                    if k not in results:
+                        results[k] = [[t, v]]
+                    else:
+                        results[k].append([t, v])
+
+            except ParseException:
+                continue
+
+        return results
+
+    def find_binary(self, ip_version, host, interval, length, target):
+        script = os.path.join(DATA_DIR, 'scripts', 'ss_iterate.sh')
+        if not os.path.exists(script):
+            raise RuntimeError("Cannot find ss_iterate.sh.")
+
+        bash = util.which('bash')
+        if not bash:
+            raise RuntimeError("Socket stats requires a Bash shell.")
+
+        resol_target = util.lookup_host(target, ip_version)[4][0]
+        if ip_version == 6:
+            resol_target = "[" + str(resol_target) + "]"
+
+        return "{bash} {script} -I {interval:.2f} " \
+                "-c {count:.0f} -H {host} -t \"{target}\"".format(
+                        bash=bash,
+                        script=script,
+                        interval=interval,
+                        count=length // interval + 1,
+                        host=host,
+                        target=resol_target)
 
 
 class TcRunner(ProcessRunner):
--- a/flent/scripts/ss_iterate.sh
+++ b/flent/scripts/ss_iterate.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+#Copyright (C) 2016  Matthias Tafelmeier
+
+#ss_iterate.sh is free software: you can redistribute it and/or modify
+#it under the terms of the GNU General Public License as published by
+#the Free Software Foundation, either version 3 of the License, or
+#(at your option) any later version.
+
+#ss_iterate.sh is distributed in the hope that it will be useful,
+#but WITHOUT ANY WARRANTY; without even the implied warranty of
+#MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#GNU General Public License for more details.
+
+#You should have received a copy of the GNU General Public License
+#along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+count=10
+interval=0.1
+host=localhost
+
+usage()
+{
+    echo "$0 -c count -I interval -H host -t target"
+}
+
+while getopts "c:I:H:t:p:" opt; do
+    case $opt in
+        c) count="$OPTARG" ;;
+        I) interval="$OPTARG" ;;
+        H) host="$OPTARG" ;;
+        t) target="$OPTARG" ;;
+    esac
+done
+
+if [ -z "$target" ]
+then
+    usage
+    exit 1
+fi
+
+command_string=$(cat <<EOF
+for i in \$(seq $count); do
+    ss -t -i -p -n state connected dst "$target"
+    date '+Time: %s.%N';
+    echo "---";
+    sleep $interval || exit 1;
+done
+EOF
+)
+
+if [[ "$host" == "localhost" ]]; then
+    eval "$command_string"
+else
+    echo $command_string | ssh $host sh
+fi


flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,50 @@
...
+        I) interval=$OPTARG ;;
+        H) host=$OPTARG ;;
+		t) target=$OPTARG ;;
+    esac
+done
+
+buffer=""
+[[ "$host" == "localhost" ]] || buffer="-b"
+
+cmd_generic="ss -t -i | grep -A 1 $target"
```
tohojo: ss has a filtering option; probably better to use that than grepping?

Also, using ss -n is probably better, and -H will keep you from having to discard the header later, and I'm guessing you're going to need -p to match process numbers to netperf instances; there are going to be several sockets per host in many cases.

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,50 @@
...
+command_string=$(cat <<EOF
+#which ss_iterate >/dev/null && exec tc_iterate $buffer -i $interface -c $count -I $interval -C $command; // not rdy yet
+for i in \$(seq $count); do
+	$cmd_generic | egrep -o '\<cwnd:[0-9]*\>' | sed 's/cwnd://g' | column;
+	$cmd_generic | egrep -o '\<rtt:[0-9]*\>' | sed 's/rtt://g' | column;
```
tohojo: Probably better just to capture everything and parse it inside Flent; also saves you from having to rely on egrep and sed being available and compatible, and if you want to add another field later, you don't need to change the script (or, more importantly, the compiled binary when such a thing exists).

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,50 @@
...
+count=10
+interval=0.1
+host=localhost
+
+while getopts "c:I:H:t:" opt; do
+    case $opt in
+        c) count=$OPTARG ;;
+        I) interval=$OPTARG ;;
+        H) host=$OPTARG ;;
+		t) target=$OPTARG ;;
```
tohojo: Indentation error

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,50 @@
...
+    esac
+done
+
+buffer=""
+[[ "$host" == "localhost" ]] || buffer="-b"
```
tohojo: Maybe just get rid of this until there's a need for it?

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,50 @@
...
+
+cmd_generic="ss -t -i | grep -A 1 $target"
+command_string=$(cat <<EOF
+#which ss_iterate >/dev/null && exec tc_iterate $buffer -i $interface -c $count -I $interval -C $command; // not rdy yet
```
tohojo: I would get rid of this as well; pretty sure having a comment in here is going to break the script when it is being piped to a remote host...

cherusk:
Sure, I am doing the runner and might more ... only left it there yesterday

tohojo:
Awesome! Looking forward :)

cherusk:
Ok, Runner is up, I've tested parsing - all fine. If the result dict is immeditaly diggestable by your transformers, I've not looked into, as a **primer**.

dtaht:
The sleep .1 will break on ash. I ended up basically rolling a C program to do this which you will find several variants of in the misc directory.

dtaht:
Darn. I was kind of hoping that the ss -E option would be useful here. That shows events as they happen but seems to be limited in my limited testing.

On the other hand it is way more lightweight than forking ss all the time. Not useful?

dtaht:
Proof of concept patch for ss polling, which is WAY more efficient than any other method, and I think an improved variant could make it upstream (ash doesn't have ss, either, it's part of the iproute tools)

God, this code is so 90s.

diff --git a/misc/ss.c b/misc/ss.c
index 4454bd1..cee8abe 100644
--- a/misc/ss.c
+++ b/misc/ss.c
@@ -4006,6 +4006,8 @@ int main(int argc, char *argv[])
        int ch;
        int state_filter = 0;
        int addrp_width, screen_width = 80;
+       int duration = 10;
+       int interval = 10;
 
        while ((ch = getopt_long(argc, argv,
                                 "dhaletuwxnro460spbEf:miA:D:F:vVzZN:KHS",
@@ -4348,7 +4350,7 @@ int main(int argc, char *argv[])
 
        if (follow_events)
                exit(handle_follow_request(&current_filter));
-
+       do {
        if (current_filter.dbs & (1<<NETLINK_DB))
                netlink_show(&current_filter);
        if (current_filter.dbs & PACKET_DBM)
@@ -4368,6 +4370,7 @@ int main(int argc, char *argv[])
 
        if (show_users || show_proc_ctx || show_sock_ctx)
                user_ent_destroy();
-
+       sleep(1);
+       } while (--duration > 0);
        return 0;
 }

cherusk:
@dtaht 

- Regarding ash: why should we be wary? It's some bsd sh derivate as I understand it. 

- About what "ss -E" are you talking about. Cannot confirm it for latest build.

- To your polling approach: That'd be viable way to go in my eyes. :) That's what I basically put on Flent users ML - altering existing tooling directly instead of wrapping it.

Though, for the time being and to get track, I'm convinced my interim solution is good enough to deliver acceptable granularity.

tohojo:
I'll get around to reviewing this in a few days...

cherusk:
A few days have passed. :)

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

> A few days have passed. :)

Yeah, sorry about that. Have been swamped, but have some time on a train
this afternoon. Will look at it there :)

-Toke

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+#Copyright (C) 2016  Matthias Tafelmeier
+
+#ss_iterate.sh is free software: you can redistribute it and/or modify
+#it under the terms of the GNU General Public License as published by
+#the Free Software Foundation, either version 3 of the License, or
+#(at your option) any later version.
+
+#lindwurm is distributed in the hope that it will be useful,
```
tohojo: lindwurm?

flent/runners.py
```diff
@@ -844,6 +844,59 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
 
+class SsRunner(ProcessRunner):
+    """Runner for iterated `ss -t -i -p`. Depends on same partitial output
+    separationa and time stamping as TcRunner."""
+
+    time_re = re.compile(r"^Time: (?P<timestamp>\d+\.\d+)", re.MULTILINE)
+    
+    CWND_IDX = 0
+    RTT_IDX = 1
+    PID_IDX = 2
```
cherusk: This positional approach made the Runner parsing code less convoluted, though I am prepared to align. See your point.

flent/scripts/ss_iterate.sh
cherusk: Gosh! Another one. You're really looking closely.

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,49 @@
...
+count=10
+interval=0.1
+host=localhost
+
+while getopts "c:I:H:t:" opt; do
+    case $opt in
+        c) count=$OPTARG ;;
+        I) interval=$OPTARG ;;
+        H) host=$OPTARG ;;
+        t) target=$OPTARG ;;
```
cherusk: Come on, have you checked that for the existing scripts? Those scripts do not seem to be in **interactive** use at all.

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,49 @@
...
+    esac
+done
+
+cmd_generic="ss -t -i -p | grep -A 1 $target"
```
cherusk: Ah, bpf based. Dear, slipped my mind, you're right. That'd save us a lot.

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,49 @@
...
+command_string=$(cat <<EOF
+#which ss_iterate >/dev/null && exec tc_iterate $buffer -i $interface -c $count -I $interval -C $command; // not rdy yet
```
cherusk: ACK

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,49 @@
...
+for i in \$(seq $count); do
+	TMP_OUT=\`$cmd_generic\`
+	echo "\$TMP_OUT" | egrep -o '\<cwnd:[0-9]*\>' | sed 's/cwnd://g' | column;
+	echo "\$TMP_OUT" | egrep -o '\<rtt:[0-9]*\>' | sed 's/rtt://g' | column;
+	echo "\$TMP_OUT" | egrep -o '\<pid=[0-9]*\>' | sed 's/pid=//g' | column;
```
cherusk: Ok, the netperf matching does muddy waters a little. From optimality perspective, I'm fully aware this is a nightmare, as already pointed out but only **INTERIM**.

Fine, am shifting format lifting into the parser.

cherusk:
Am not purposing to challenge flent design at all and that's set now: some orchestration tool like **Ansible** would've been convenient for your gathering backbone, as it appears to me. 

However, am commencing on that asap and trying to cover the tying up. Might need your involvement there then.

Thanks for your reviewing spent time. 

Cheers.

flent/runners.py
```diff
@@ -844,6 +844,59 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+            except ValueError:
+                pass
+
+            for cwnd,rtt,pid in map(None, \
+                    raw_lines[CWND_IDX].split(), raw_lines[RTT_IDX].split(), raw_lines[PID_IDX].split()):
+                if pid:
+                    if pid not in results:
+                        results[str(pid)] = [[timestamp, cwnd, rtt]]
+                    else:
+                        results[str(pid)].append([timestamp, cwnd, rtt])
```
cherusk: ACK

flent/runners.py
```diff
@@ -844,6 +844,59 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+        parts = output.split("\n---\n")
+        for part in parts:
+            timestamp = time_re.search(part)
+            if timestamp is None:
+                continue
+            timestamp = float(timestamp.group('timestamp'))
+
+            raw_lines = part.splitlines()
+            try:
+                raw_lines.remove('')
```
cherusk: Indeed, aux mechanism for following map ...

cherusk:
@tohojo 
Have aligned the existing parts.

Regarding the runner:
It surfaced, that ss is swallowing the PID flag -p in combination with filtering. Haven't checked it in ss upstream yet. Appears as a bug to me.

I'd prefer to stick with ss filtering, though. As a consequence, I'd suggest doing tying up with netperf based on the **five flow tuple** or a sub part of it. Should be unique enough, although for there being OS specialities around. Checked superficially in your cnfg by greping around and on the netperf page. That should be supported from netperf side, if not mistaken. 

Kindly, appreciate your opinion on that.

BR

flent/scripts/ss_iterate.sh
tohojo: They're not in interactive use, no, but they have defaults for all parameters. 

What I want to achieve here is that it should be obvious why the script fails if it does. And if it's just outputting grep errors, that is not obvious. I think that is less of a concern with the ss filtering, but I'd like to be fairly certain that a debug log (which contains the command output prior to parsing) gives a useful clue if something breaks...

tohojo:
Getting there! Will go through it in detail later.

Are you sure the missing pid is not a permission issue? If you're not root, ss will only show pids for sockets owned by the same user that is running ss itself. For our use case that shouldn't matter?

How do you propose getting the port number from netperf? Do you want to tell netperf to use a specific port number? If so, what happens if that port is not available for whatever reason?

flent/scripts/ss_iterate.sh
cherusk: Fair enough, let me introduce a usage predicate. Swallowing an err by defaults here is even more a dread in the direction of debugging.

cherusk:
> Are you sure the missing pid is not a permission issue [...]
Got me there, it actually is.  Hm , excuses for shooting before checking.

> For our use case that shouldn't matter?
Not, at all. We can stick with the pid approach then, I'd say.

>Do you want to tell netperf to use a specific port number?  
Yes, that was my intention. I thought of a randomization or a counter mechanism with checks as a last resort. The port space should be no concern. However, let's stick with the pid way and let's preserve the flow specs idea in the back of our heads. It may does add up later.

tohojo:
Cool. In terms of api between the netperf and ss runners, I was thinking that we will probably always want the ss data whenever we're running a netperf instance. So maybe the ss runner should be a child runner of the netperf runner, conceptually? Then, the netperf runner can chain the start of the ss runner, as well as the parsing step (and supply a pid for the latter).

Does that make sense?

cherusk:
Sounds reasonable enough and that was seeping through to me from your previous statements. One point: we've ss-runner knows the PID in advance, then we'd do the filtering on that in the iterator script again on top of ss filtering. Minor one, one additional fork should be fine. Hm ... what confuses me why I've to parse the flow endpoints (ports) then. There is one netperf client instance per flow, right?

tohojo:
Well, you won't necessarily know the PID ahead of time (i.e. early
enough to filter it). The ss and netperf instances are going to start
roughly at the same time, so to be sure we have the pid, SS would have
to wait, which may mean it misses something. However, at the time of
parsing (once the runners are done), we'll certainly have the pid, so
filtering there should be safe.

And you'll need the ports because netperf opens a control connection to
netserver that is separate from the data connection. You'll need to
filter that out. But that's on a known port (12865), so should be simple
enough to filter out...

-Toke

cherusk:
I'm relieved not having implemented something superfluous, then. FYI - Doing the ports parsing atm ...

cherusk:
Ok, now all the data is at least in flent, now, I'd say. 

Kindly over to you, now.

flent/runners.py
```diff
@@ -845,6 +846,95 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+class ParseException(Exception):
+    pass
+
+
+class SsRunner(ProcessRunner):
+    """Runner for iterated `ss -t -i -p`. Depends on same partitial output
+    separationa and time stamping as TcRunner."""
+
+    ip_v4_addr_sub_re = "(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])"
+    ip_v6_addr_sub_re = "(?:(?:[A-F0-9]{1,4}:){7}[A-F0-9]{1,4}|(?=(?:[A-F0-9]{0,4}:){0,7}[A-F0-9]{0,4}(?![:.\w]))(([0-9A-F]{1,4}:){1,7}|:)((:[0-9A-F]{1,4}){1,7}|:))(:\d+)+"
```
tohojo: While I can appreciate the preciseness of these, I fear they are going to be hard to debug going forward. I wonder if it can be simplified; for instance having something like "([0-9]{1,3}\.){3}[0-9]{1,3}" match an IPv4 address would be fine; the risk of having addresses >255 show up is not that high.

But, well, on the other hand I can live with those if you prefer them. But in that case, please at least comment the different parts of the IPv6 regex so it is readable to those with less regex foo (I'm not sure I get the different parts myself of that one).

flent/runners.py
```diff
@@ -845,6 +846,95 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+
+    time_re = re.compile(r"^Time: (?P<timestamp>\d+\.\d+)", re.MULTILINE)
+    cwnd_re = re.compile(r"cwnd:(?P<cwnd>\d+)", re.MULTILINE)
+    rtt_re = re.compile(r"rtt:(?P<rtt>\d+.\d+/\d+.\d+)", re.MULTILINE)
```
tohojo: You should probably escape the . here

flent/runners.py
```diff
@@ -845,6 +846,95 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+    ports_ipv4_re = re.compile(r"" + ip_v4_addr_sub_re +
+                               ":(?P<src_p>\d+)\s+?" + ip_v4_addr_sub_re + ":(?P<dst_p>\d+)")
+    ports_ipv6_re = re.compile(r"" + "(?P<src_p>" + ip_v6_addr_sub_re + ")" + "\s+" +
+                               "(?P<dst_p>" + ip_v6_addr_sub_re + ")", re.IGNORECASE)
+
+    src_p = []
+    dst_p = []
+    pid = []
+
+    def parse_part(self, part):
```
tohojo: Currently, this will only parse the first line in a "part" - so if we have multiple sockets to the same destination (which we are going to have), it will fail to find anything but the first. You'll need to loop through all lines in an output part and through out anything that's not related to the pid you're looking for.

How to get the pid is the inter-runner API issue we were talking about. I can look into stubbing something out in the base class, unless you want to give it a go yourself?

flent/runners.py
```diff
@@ -845,6 +846,95 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+            raise ParseException()
+
+        timestamp = float(timestamp.group('timestamp'))
+        cwnd = int(cwnd.group('cwnd'))
+        rtt, rtt_var = [float(x) for x in rtt.group('rtt').split('/')]
+        pid = int(pid.group('pid'))
+        self.src_p.append(f_ports.group('src_p').split(":")[-1])
+        self.dst_p.append(f_ports.group('dst_p').split(":")[-1])
+
+        val_res = {'cwnd': cwnd, 'rtt': rtt, 'rtt_var': rtt_var}
```
tohojo: If you stick the timestamp in here with a 't' key, you'll basically have the dict you need for raw_values (see below).

flent/runners.py
```diff
@@ -845,6 +846,95 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+            try:
+                t, res_dict, pid = self.parse_part(part)
+                for k, v in res_dict.items():
+                    if k not in results:
+                        results[k] = [[t, v]]
+                    else:
+                        results[k].append([t, v])
+
+                self.pid.append(pid)
+                raw_values = copy.deepcopy(results)
```
tohojo: Raw values should basically be what is currently res_dict, but with the timestamp as the 't' key. No need to copy (and see suggestion above about just having parse_part return the raw_values dict).

flent/runners.py
```diff
@@ -845,6 +846,95 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+            raise RuntimeError("Socket stats requires a Bash shell.")
+
+        return "{bash} {script} -I {interval:.2f} " \
+            "-c {count:.0f} -H {host} -t {target}".format(
+                bash=bash,
+                script=script,
+                interval=interval,
+                count=length // interval + 1,
+                host=host,
+                target=target,
```
tohojo: On my machine, at least, ss will not work correctly when filtering on a host name. So we'll need to make sure that target is a suitable IP address, and if it's an IPv6 address, that it is enclosed in [] (and escaped).

flent/runners.py
```diff
@@ -845,6 +846,95 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+                port=port)
```
tohojo: Why is the port still there?

flent/runners.py
```diff
@@ -34,6 +34,7 @@
 import tempfile
 import threading
 import time
+import copy
```
tohojo: Not needed, I think (see below).

flent/runners.py
cherusk: Lapsus.

flent/runners.py
cherusk: And restricting it to IP only? 

Haaa? ss really requires that on IPv6 address! Hm ... hopefully, I've already complained about the current state of ss in your presence.

Well, either or, escaping or [] is necessary as I understand it from file:///usr/share/doc/iproute2-doc/ss.html. Not both.

Mnor one ...

flent/runners.py
cherusk: Thanks.

flent/runners.py
tohojo: Yeah, according to the docs you can use a hostname. According to my tests, that produces no output... :/

Flent already does some hostname lookup somewhere, but I forget whether it'll use the numeric addresses when passing it to the runners...

(And yes, I agree about the state of ss; but such is life)

flent/runners.py
cherusk: Indeed, thanks.

flent/runners.py
cherusk: Although I could, I've not completely crafted them on my own, since of being reluctant to reinventing commonplaces. Let's keep them and I'll add the reference where one can find the commented version of the ipv6_addr RE.

flent/runners.py
cherusk: Right, was still not doing anything according the filtering since of the open API quarrel. Hm, you know your way around better, I know where runners are kicked off, though I'd have to start a grep-marathon still. Thus, I'd appreciate you delineating the framing.

flent/runners.py
cherusk: Ah, ok, doing that from python flent. Would have been my next question. 

Well, good you found that one. 

> (And yes, I agree about the state of ss; but such is life)
Well, agreeing ... for the time being.

cherusk: Further, I did not get ... are you covering that one?

tohojo: I'll go lookup what we do to the hostname in other case and think about the best way to do it in this case... :)

tohojo:
Okay, added the API bits to be able to add child runners.

There are two changes to the behaviour of runners that you'll need to leverage:

One is that there is now an add_child() method in BaseRunner; you can call this as appropriate to add a child runner to the netperf runner. So in this instance, call it from find_binary in the netperf runner for the case where we're running a TCP flow. You give add_child a class (your ss runner class) and any positional args that this runner needs for its constructor (so the positional args to the ProcessRunner constructor in this case). Child runners will have their parent as the _parent attribute, which you can use to get the pid from the process runner at parse time.

The second change is that find_binary can now be an instance method instead of a class method. If the 'commands' arg to a ProcessRunner is a dictionary, it will fall find_binary at runner init time, with the commands dictionary as kwargs. If you combine this with the above, you should be able to instantiate your runner with the right args.

Finally, there is a lookup_host function in the util module that you can use to lookup the target host name (see example in the find_binary method of the ping runner (line 824 after the latest updates)). And if you pass in the IP version, adding the [] escape is trivial.

The only thing missing is figuring out what to do with the results of the child runner. But I figure that can wait until you have the parser working; there should be enough functionality there now for you to get it running and the parser working...

Let me know if the above doesn't make sense, or if there's something else I missed... ;)

cherusk:
It felt mostly elegant. I'd say we're good to go now. Appreciating you coming back so quickly.

What I felt about odd when preparing args for the kwargs unroll to my find_binary in netperf runner, but let's assume rare API changes.

flent/runners.py
```diff
@@ -680,8 +680,14 @@ def find_binary(self, **args):
             args['socket_timeout'] = ""
 
         if args['test'] in ("TCP_STREAM", "TCP_MAERTS", "omni"):
+            ss_args = {}
+            for k, v in args:
+                if k in ["host", "interval", "length", "target", "ip_version"]:
+                    ss_args[k] = v
```
tohojo: Hmm, looping over everything in args is somewhat inefficient. Also, the parameters are not called the same, so you'll need to map them. Using a dict literal is probably the easiest. Something like:

```
ss_args = {'host': self.remote_host or 'localhost', 
           'interval': args['interval'],
           'length': args['length'],
           'target': args['host'],
           'ip_version': args['ip_version']}
````

File: flent/runners.py [hunk 23]
```diff
@@ -680,8 +680,14 @@ def find_binary(self, **args):
...
+            self.add_child(SsRunner, ss_args)
```
tohojo: You'll need to also pass the two other positional arguments to ProcessRunner (which is where the init method that will be called for SsRunner lives). Delay is in `self.delay` and `remote_host` should be None (because that is passed as the 'host' arg above).

flent/runners.py
```diff
@@ -950,6 +956,126 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+        for sp in sub_parts:
+            pid = self.pid_re.search(part)
+            if None is pid:
+                continue
+            pid_str = pid.group('pid')
+            if self.pid == pid_str:
+                sub_part.append(sp)
+
+        if 1 != len(sub_part):
+            raise ParseException()
```
tohojo: This is going to trigger if the same process has more than one socket open to the same target. Which netperf will have in most cases (one for the control and one for the data connection). So I think you'll need to move the port number parsing to here and filter on the control connection port.

flent/runners.py
```diff
@@ -950,6 +956,126 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+        bash = util.which('bash')
+        if not bash:
+            raise RuntimeError("Socket stats requires a Bash shell.")
+
+        resol_target = util.lookup_host(target, ip_version)[4][0]
+        if ip_version == 6:
+            resol_target = "[" + str(resol_target) + "]"
+
+        return "{bash} {script} -I {interval:.2f} " \
+            "-c {count:.0f} -H {host} -t {target}".format(
```
tohojo: You'll need to single-quote target, otherwise the shell is going to barf on the []

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,55 @@
...
+
+if [ -z "$target" ]
+then
+    usage
+    exit 1
+fi
+
+command_string=$(cat <<EOF
+for i in \$(seq $count); do
+    ss -t -i -p -n state connected dst $target
```
tohojo: I think target is going to need quoting here as well, for the same reason as above.

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,55 @@
...
+host=localhost
+
+usage()
+{
+    echo "$0 -c count -I interval -H host -t target"
+}
+
+while getopts "c:I:H:t:p:" opt; do
+    case $opt in
+        c) count=$OPTARG ;;
```
tohojo: Probably a good idea to quote $OPTARG in all these; target at least is likely to need it...

flent/runners.py
```diff
@@ -950,6 +956,126 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+class ParseException(Exception):
+    pass
+
+
+class SsRunner(ProcessRunner):
+    """Runner for iterated `ss -t -i -p`. Depends on same partitial output
+    separationa and time stamping as TcRunner."""
+
+    ip_v4_addr_sub_re = "(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])"
+    # ref.: to commented, untinkered version: ISBN 978-0-596-52068-7
```
tohojo: What is the license of the code examples in that book?

flent/runners.py
```diff
@@ -950,6 +956,126 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+
+        return sub_part[0]
+
+    def parse_part(self, part):
+
+        sub_part = self.filter_np_parent(part)
+
+        timestamp = self.time_re.search(sub_part)
+        cwnd = self.cwnd_re.search(sub_part)
+        rtt = self.rtt_re.search(sub_part)
```
tohojo: Should probably check if any of these fail; otherwise the code below will crash...

flent/runners.py
cherusk: > Using Code Examples This book is here to help you get your job done. In general, you may use the code in this book in your programs and documentation. You do not need to contact us for permission unless you’re reproducing a significant portion of the code. For example, writing a program that uses several chunks of code from this book does not require permission. Selling or distributing a CD-ROM of examples from O’Reilly books does require permission. Answering a question by citing this book and quoting example code does not require permission. Incorporating a significant amount of example code from this book into your product’s documentation does require permission. We appreciate, but do not require, attribution. An attribution usually includes the title, author, publisher, and ISBN. For example: “Regular Expressions Cookbook by Jan Goyvaerts and Steven Levithan. Copyright 2009 Jan Goyvaerts and Steven Levithan, 978-0-596-2068-7.”

So, I'd might extend the ref. ;)

flent/runners.py
cherusk: I've tested all introduced code paths, but not the integrated thing. Thought of runners unit tests ...

cherusk:
Ok, let's proceed ... :)

flent/runners.py
tohojo: Heh, that's not a license. Ah well, guess a one-liner falls under what is allowed...

flent/runners.py
tohojo: But what happens if the system ss version doesn't include one of the keys?

flent/scripts/ss_iterate.sh
```diff
@@ -0,0 +1,55 @@
...
+host=localhost
+
+usage()
+{
+    echo "$0 -c count -I interval -H host -t target"
+}
+
+while getopts "c:I:H:t:p:" opt; do
+    case $opt in
+        c) count='$OPTARG' ;;
```
tohojo: Single quotes prevent variable expansion in shell...

flent/runners.py
cherusk: For most enterprise/non-enterprise distros I'd say this one is to neglect. It's been around for ~15 y and netdev is quite considerate about not breaking "user land" (know ss is user land). And I've tested it for most recent distros ... further, if so, we'll fix it ^^ More?

flent/runners.py
```diff
@@ -950,55 +959,180 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+                        script=script,
+                        interval=interval,
+                        count=length // interval + 1,
+                        host=host,
+                        target=resol_target)
+
+
 class TcRunner(ProcessRunner):
-    """Runner for iterated `tc -s qdisc`. Expects iterations to be separated by
+                    """Runner for iterated `tc -s qdisc`. Expects iterations to be separated by
```
cherusk: Vim key combination accident ... cured it. Need to praise **git add -p** once more.

cherusk:
Did update commit msgs. Am familiar with kernel dev.

As a side not, you're not signing off your commits. :)

flent/runners.py
tohojo: Just skipping the entry if the regexes don't match (as you do now) is fine. Empty results when an error happens is OK. Crashing, not so much ;)

As for unittests, there's a single one for the Cake parser in unittests/test_parsers.py - because that kept breaking. Have not added anything else so far; but feel free to add one for this if you want :)

flent/runners.py
```diff
@@ -680,8 +680,17 @@ def find_binary(self, **args):
...
 
         if args['test'] in ("TCP_STREAM", "TCP_MAERTS", "omni"):
+            self.ctrl_port = args['netperf_control_port']
+            ss_args = {'host': self.remote_host or 'localhost',
+                       'interval': args['interval'],
+                       'length': args['length'],
+                       'target': args['host'],
+                       'ip_version': args['ip_version']}
+
+            self.add_child(SsRunner, self.delay, None, ss_args)
```
tohojo: This needs to match the order of the ProcessRunner args, so ss_args goes before delay.

tohojo:
Yeah, I know, but so far I've been the only developer, so not much need. But you're right, I guess I should start doing that...

Apart from the one nit with command args I commented on above, I can't see anything else that's missing. Unfortunately, I did not have time to test it today, will try to get around to it tomorrow...

flent/runners.py
cherusk: Hold on: 

>  ProcessRunner:
>    def __init__(self, command, delay, remote_host, **kwargs):

I'd say I'm doing it correctly.

cherusk:
Don't mind you. 

Do your tests. Not sure what is harrassing you regarding the interpretation of the results - a simple time serial depiction should do it. Anyways, will learn it soon, I assume.

flent/runners.py
tohojo: No, ss_args is the 'command' of the processrunner. That is overloaded, so that a dict passed as that parameter will be expanded into kwargs for the find_binary function.

cherusk: Ah, overlooked that in my reaction ...

cherusk: Up!

flent/runners.py
```diff
@@ -680,8 +680,17 @@ def find_binary(self, **args):
             args['socket_timeout'] = ""
 
         if args['test'] in ("TCP_STREAM", "TCP_MAERTS", "omni"):
+            self.ctrl_port = args['netperf_control_port']
```
tohojo: There's no 'netperf_control_port' in args, so this just crashes.

We don't actually configure the control port anywhere, so you'll just have to hard-code it somewhere. Maybe making ssrunner more generic by having a notion of "excluded ports" and passing in 12865 in an argument to the constructor or something?

flent/runners.py
```diff
@@ -950,6 +959,131 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+    src_p = []
+    dst_p = []
+    # ref.: upstream ss
+    ss_states = ["UNKNOWN", "ESTAB", "SYN-SENT", "SYN-RECV", "FIN-WAIT-1",
+                 "FIN-WAIT-2", "TIME-WAIT", "UNCONN", "CLOSE-WAIT", "LAST-ACK",
+                 "LISTEN", "CLOSING"]
+    ss_states_re_pat = "|".join(ss_states)
+
+    def __init__(self, **kwargs):
+        super(SsRunner, self).__init__(**kwargs)
```
tohojo: You'll need to accept positional arguments (*args) in the constructor and pass then on to the parent as well.

flent/runners.py
```diff
@@ -950,6 +959,131 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+        if 1 != len(sub_part):
+            raise ParseException()
+
+        return sub_part[0]
+
+    def parse_part(self, part):
+
+        sub_part = self.filter_np_parent(part)
+
+        timestamp = self.time_re.search(sub_part)
```
tohojo: You're not going to find the timestamp in sub_part; that should probably be 'part' :)

flent/runners.py
cherusk: Hold on:
Doing from flent

> $ grep -iR NETPERF_CONTROL_PORT *
> flent/settings.py:    action="store", type=int, dest="NETPERF_CONTROL_PORT", metavar="PORT",

You have it as a default in your settings, remember? Checked for other params that are in settings.py. You draw them from args dict as well.

cherusk:
Could you kindly retest, please? Hope, that was all.

flent/runners.py
tohojo: Huh, whatdoyaknow, so it is. It's not used, though, and so never propagated to the args dict. Guess I added the command line parameter at some point, but never got around to actually using it...

tohojo: (and yeah, guess I'll go fix that ;))

cherusk: Is it tough to propagate it? Appears cleanest to me. Otherwise ...

tohojo: Nah, just pushed the change; it's called 'control_port' in the args dict, though, to be consistent with the existing 'control_host' :)

cherusk: Cool, just a jiffy.

flent/runners.py
```diff
@@ -950,6 +959,132 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+    src_p = []
+    dst_p = []
+    # ref.: upstream ss
+    ss_states = ["UNKNOWN", "ESTAB", "SYN-SENT", "SYN-RECV", "FIN-WAIT-1",
+                 "FIN-WAIT-2", "TIME-WAIT", "UNCONN", "CLOSE-WAIT", "LAST-ACK",
+                 "LISTEN", "CLOSING"]
+    ss_states_re_pat = "|".join(ss_states)
+
+    def __init__(self, *args):
+        super(SsRunner, self).__init__(*args)
```
tohojo: Ah, no, you need both *args and **kwargs :)

tohojo:
This is what I'm doing to test: `./run-flent -v -l 4 --debug -4 tcp_1up netperf-eu.bufferbloat.net` - please try running that yourself and make sure it doesn't produce any parse errors... :)

cherusk:
Looks good on my end:
[...]

Stderr: 
NetperfDemoRunner TCP upload finished
PingRunner Ping (ms) ICMP finished
Runner aggregation finished
Data file written to ./tcp_1up-2017-02-18T205717.043897.flent.gz.
Creating new DefaultFormatter
Summary of tcp_1up test run at 2017-02-18 19:57:17.043897:

                           avg       median          # data pts
 Ping (ms) ICMP :       146.66        63.68 ms               70
 TCP upload     :         1.36         1.57 Mbits/s          21

**Maybe, something in your water. ;)**

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

> Looks good on my end:
> [...]
>
> Stderr:

What did you skip here?

-Toke

cherusk:
Only the ss_iterate sc output. Ah, wait ...

Here - ok, gosh!

SsRunner TCP upload :: child 0 finished

> WARNING: Command produced no valid data.

Runner class: SsRunner

tohojo:
Exactly ;) the output only appears because there's no parsed data (as a debug message). You won't see the data show up anywhere yet (I'll go fix that next), so the only indication that it fails is that error...

cherusk:
Aehm, tracked it down: There appear to be two "runs".
----
{}
False
----
[...]
----
[[1487449410.491, 2.17], [1487449410.766, 1.91], [1487449411.011, 1.6], [1487449411.353, 1.53], [1487449411.721, 1.43], [1487449412.112, 1.67], [1487449412.318, 1.27], [1487449412.724, 1.61], [1487449412.953, 1.15], [1487449413.167, 1.83], [1487449413.417, 1.05], [1487449413.646, 1.72], [1487449413.905, 1.52], [1487449414.16, 1.54], [1487449414.25, 2.93]]
False
----


One ends correctly, the other is empty. Remember you saying someting that there might be redundant netperf runs.

So nothing in the ss parser per se. 

Appears this predicate needs adaption:

> **if not self.result and not self.silent:** (448)

tohojo:
Huh, that's odd. It shouldn't actually go through run more than once.

Have prototyped the code that integrates the data into the data file; let me see what it looks like with that...

tohojo:
Actually, that looks like the second one that you printed is the netperf runner itself (I assume you added a print in processrunner.run()?). Which would make the first one the output from the ss parser, which is empty...

tohojo:
To debug, I'll suggest you print the result of the SsRunner parse() method just before it returns, and make sure that looks right :)

And yeah, the runner should probably be marked as silent eventually; or we should make all child runners silent. But for now, the error is useful for debugging :)

cherusk:
Yeah, am tracking with **logger.warning**

Give me a sec ...

cherusk:
Ok, tracing in **filter_np_parent**:

WARNING: port---
WARNING: 42912
WARNING: 12865
WARNING: True
WARNING: pid---
**WARNING: 31831
WARNING: 31832**
WARNING: False
WARNING: port---
WARNING: 12865
WARNING: 12865
WARNING: False
WARNING: pid---
**WARNING: 31831
WARNING: 31832**
WARNING: False

Therefore, self._parent.pid always off 1 to actual netperf instance pid reported by ss. Huh?

tohojo:
Hmm, that is somewhat odd. What does the log output say? The "Forked /usr/bin/netperf as pid XXX" line...

cherusk:
Aligns with ss, as expected:

Forked /usr/bin/netperf as pid 31831

tohojo:
Ah, hang on... I know what's going on.

You are using self.pid to store the parent pid. But when the ss collection script itself is run afterwards, that will set self.pid to the pid of the script; which is one pid higher.

So you'll need to store it as self.parent_pid or something and use that when checking :)

cherusk:
Have it.

Just FYI:

doing in SsRunner __init__

self.pid = str(self._parent.pid)
But self._parent.pid seems not available by then, ever fetching after super initilization. Some python Constructor precedence. 

So self.pid afterwards is that of SsRunner itself.

cherusk:
Race condition ...

tohojo:
Not really a race condition; just a namespace collision causing the superclass to override self.pid... (see my simultaneous comment above)

cherusk:
Meant our findings were racing!

tohojo:
Ah, right; yes, nice timing on that one ;)

cherusk:
Odd is, that is working fine:
self.ctrl_port = str(self._parent.ctrl_port)

Anyways!

tohojo:
The pid instance variable is used by ProcessRunner, which is a parent of SsRunner. ctrl_port is not...

cherusk:
Nah, was rather in SsRunner __init self._parent.pid it's oddly None and ctrl_port not, but nevermind ...

tohojo:
Ah, right. Yes, that would be because at the time SsRunner.__init__ is run, netperf hasn't started yet, and so its pid is not known. So storing the pid in __init__ is not going to work. Silly me, should have realised that before...

cherusk:
Does not add, easily circumventable ...

Now it's working, at least this result warning is gone:

q.e.d.:
Executing test environment file /home/matthias/Projects/sync/cc_qdisc/flent/flent/tests/tcp_1up.conf
Started Flent 1.0.1-git-157e70d using Python 2.7.12.
Executing test environment file /home/matthias/Projects/sync/cc_qdisc/flent/flent/tests/tcp_1up.conf
Starting tcp_1up test. Expected run time: 14 seconds.
Forked /bin/ping as pid 1957
Started PingRunner idx 1 ('Ping (ms) ICMP')
TCP upload: Adding child SsRunner
Forked /usr/bin/netperf as pid 1961
Forked /bin/bash as pid 1962
Started SsRunner idx 0 ('TCP upload :: child 0')
Started NetperfDemoRunner idx 0 ('TCP upload')
SsRunner TCP upload :: child 0 finished
NetperfDemoRunner TCP upload finished
PingRunner Ping (ms) ICMP finished
Runner aggregation finished
Data file written to ./tcp_1up-2017-02-18T231906.111358.flent.gz.
Creating new DefaultFormatter
Summary of tcp_1up test run at 2017-02-18 22:19:06.111358:

                           avg       median          # data pts
 Ping (ms) ICMP :       145.74        62.23 ms               70
 TCP upload     :         1.37         1.56 Mbits/s          21

tohojo:
Working better, certainly ;)

I just ran it three times, of which it succeeded the two. And when it does succeed it only produces a single data point...

cherusk:
Ok, there are still crashes, but I challenge the single data point story at successful run.

logger.warning("%s" % results) confirms valid dict at success.

flent/runners.py
```diff
@@ -952,6 +961,132 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+        super(SsRunner, self).__init__(*args, **kwargs)
+        self.ctrl_port = str(self._parent.ctrl_port)
+
+    def filter_np_parent(self, part):
+        sub_part = []
+        sub_parts = re.split(self.ss_states_re_pat, part)
+        sub_parts = [sp for sp in sub_parts if sp.strip() and not self.ss_header_re.search(sp)]
+
+        for sp in sub_parts:
+            pid = self.pid_re.search(part)
```
tohojo: Here's the problem; this is matching on the wrong variable; should be sp, not part.

flent/runners.py
```diff
@@ -952,6 +961,132 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+            if None is pid:
+                continue
+            pid_str = pid.group('pid')
+
+            f_ports = self.ports_ipv4_re.search(sp)
+            if None is f_ports:
+                f_ports = self.ports_ipv6_re.search(sp)
+
+            if None is f_ports:
+                raise ParseException()
```
tohojo: If you just continue here instead of raising an exception, you don't need the filtering for the header above...

flent/runners.py
```diff
@@ -952,6 +961,132 @@ def find_binary(cls, host, interval, length, ip_version, local_bind=None,
...
+                               "(?P<dst_p>" + ip_v6_addr_sub_re + ")", re.IGNORECASE)
+    ss_header_re = re.compile(r"" + "State\s+Recv-Q\s+Send-Q\s+Local")
+
+    src_p = []
+    dst_p = []
+    # ref.: upstream ss
+    ss_states = ["UNKNOWN", "ESTAB", "SYN-SENT", "SYN-RECV", "FIN-WAIT-1",
+                 "FIN-WAIT-2", "TIME-WAIT", "UNCONN", "CLOSE-WAIT", "LAST-ACK",
+                 "LISTEN", "CLOSING"]
+    ss_states_re_pat = "|".join(ss_states)
```
tohojo: Why are you not compiling this regex? (compiled regexes also have a split method)

cherusk: Could not find the split for compiled re so thought non-existent. Agree here.

flent/runners.py
cherusk: Nah, want to keep that since of the res for the IPs.

cherusk:
What's next?

flent/runners.py
tohojo: Well, it also has the effect that if you have a valid subpart followed by an invalid one, you're going to discard the valid one (because you'll raise an exception on the second run through the loop). Not sure this is going to happen in practice, though, since you only get this far if the pid regex matches...

tohojo:
This is mergeable now, I think. Nice job! :)

I have a changeset in my local tree that actually merges the data into the datafile, but I want to be sure that the way I'm doing it is not going to create problems (since once I push it, Flent will start producing data files in this format).

So once I've convinced myself that this is the right way to do things, I'll push those changes along with your patches.

In the meantime, it would be helpful if you could clean up these flake8 errors introduced by your patch:

```
./flent/runners.py:716:9: E303 too many blank lines (2)
./flent/runners.py:994:83: E501 line too long (161 > 82 characters)
./flent/runners.py:1000:83: E501 line too long (85 > 82 characters)
./flent/runners.py:1002:83: E501 line too long (85 > 82 characters)
./flent/runners.py:1003:83: E501 line too long (85 > 82 characters)
./flent/runners.py:1021:83: E501 line too long (95 > 82 characters)
./flent/runners.py:1061:83: E501 line too long (83 > 82 characters)
./flent/runners.py:1101:17: E127 continuation line over-indented for visual indent
```

(yeah, there's a few more errors in other files, don't worry about those)

flent/runners.py
cherusk: Not really, Toke, think twice. It's only raised if no ports are present at all. If that's the case, we only learn flent and ss have become disaligned.

cherusk: So, it's really an Exception in my eyes ...

cherusk:
Thanks for your patience and reviewing effort.

There's a german saying along those lines:
> What does last long does tend to end well. :)

tohojo:
Right, merged, with a few updates. Also added plots of cwnd and TCP RTT.

You'll need to add the --socket-stats switch to enable the capturing for now; there are some performance issues when there are a lot of simultaneous flows, which makes it unsuitable for being enabled by default.

dtaht:
yea! The love here was just AWESOME... I was dancing...

Until I clicked on the tcp_rtt_cdf plot.
[rrul-2017-02-19T122712.250266.socketstats_first.flent.gz](https://github.com/tohojo/flent/files/786168/rrul-2017-02-19T122712.250266.socketstats_first.flent.gz)


WARNING: Falling back to Qt4 for the GUI. Please consider installing PyQt5.

Initialised matplotlib v1.5.1 on numpy v1.11.0.
GUI loaded. Running on PyQt v4.11.4.
ERROR: Aborting plotting due to error: list index out of range
ERROR: Aborting plotting due to error: list index out of range

tohojo:
Dave Täht <notifications@github.com> writes:

> yea! The love here was just AWESOME... I was dancing...
>
> Until I clicked on the tcp_rtt_cdf plot.
> rrul-2017-02-19T122712.250266.socketstats_first.flent.gz
>
> WARNING: Falling back to Qt4 for the GUI. Please consider installing PyQt5.
>
> Initialised matplotlib v1.5.1 on numpy v1.11.0.
> GUI loaded. Running on PyQt v4.11.4.
> ERROR: Aborting plotting due to error: list index out of range
> ERROR: Aborting plotting due to error: list index out of range

A debug log would be useful...

-Toke

dtaht:
[rrul-2017-02-19T123738.369219.for_toke_with_love.flent.gz](https://github.com/tohojo/flent/files/786176/rrul-2017-02-19T123738.369219.for_toke_with_love.flent.gz)

dtaht:
[because_toke_loves_debug_logs.log.gz](https://github.com/tohojo/flent/files/786178/because_toke_loves_debug_logs.log.gz)

dtaht:
@cherusk thank you for this!!! I have been staring at xplot output for far far too long......

cherusk:
@dtaht 
Welcome.

Can imagine this has been reinvented for many times.

cherusk:
Now it's open ... :)

dtaht:
Well, I am now thinking that coaxing netperf to emit tcp_info would be best.

cherusk:
Hm, it will have to go via NL as well, like ss.

dtaht:
Preliminary netperf patch here: http://www.taht.net/~d/interim_netperf_tcp_info.patch

It is VERY jittery, and I don't know why. I should go and look at the quality of the ss data also.

bpf + netlink maybe?

cherusk:
At least that is what ss is doing.

dtaht:
Rick got back to me:

"Unless you use a negative value for the interval netperf will guesstimate how many sends per interval there should be, and won't check the time until that many have taken place.  If you use a negative value for the interval, it will check the time on each send (or whatever, based on the test type) and so there will likely be less jitter in the size of the actual intervals."

I wonder how much jitter is in the netperf data we have in the first place? measuring sampling jitter on our datasets might be worth while, instead of just assuming all the smoothing is smooth...

Anyway, assuming that that fixes it, we can pull this out on a per flow basis.

Interim result:    4.54 10^6bits/s over 0.202 seconds ending at 1487619256.976
tcpi_rto 225000 tcpi_ato 0 tcpi_pmtu 1500 tcpi_rcv_ssthresh 28800
tcpi_rtt 24969 tcpi_rttvar 3271 tcpi_snd_ssthresh 10 tpci_snd_cwnd 12
tcpi_reordering 4 tcpi_total_retrans 38
Interim result:    4.54 10^6bits/s over 0.231 seconds ending at 1487619257.207
tcpi_rto 232000 tcpi_ato 0 tcpi_pmtu 1500 tcpi_rcv_ssthresh 28800
tcpi_rtt 31606 tcpi_rttvar 3066 tcpi_snd_ssthresh 11 tpci_snd_cwnd 11
tcpi_reordering 4 tcpi_total_retrans 40
Interim result:    5.79 10^6bits/s over 0.294 seconds ending at 1487619257.501
tcpi_rto 233000 tcpi_ato 0 tcpi_pmtu 1500 tcpi_rcv_ssthresh 28800
tcpi_rtt 32667 tcpi_rttvar 4600 tcpi_snd_ssthresh 11 tpci_snd_cwnd 16
tcpi_reordering 4 tcpi_total_retrans 40
^CInterim result:    4.13 10^6bits/s over 0.095 seconds ending at 1487619257.596
tcpi_rto 231000 tcpi_ato 0 tcpi_pmtu 1500 tcpi_rcv_ssthresh 28800
tcpi_rtt 30749 tcpi_rttvar 3367 tcpi_snd_ssthresh 11 tpci_snd_cwnd 12
tcpi_reordering 4 tcpi_total_retrans 41

dtaht:
from iproute2 for BBR support....

commit 2f579872fbdca52ef960eb969655adfd55c13308
Author: Neal Cardwell <ncardwell@google.com>
Date:   Thu Dec 1 13:21:39 2016 -0500

    ss: print new tcp_info fields: delivery_rate and app_limited
    
    Dump the new delivery_rate and delivery_rate_app_limited fields that
    were added to tcp_info in Linux v4.9.
    
    Example output:
      pacing_rate 65.7Mbps delivery_rate 62.9Mbps
    
    And for the application-limited case this looks like:
      pacing_rate 1031.1Mbps delivery_rate 87.4Mbps app_limited
    
    Signed-off-by: Neal Cardwell <ncardwell@google.com>
    Signed-off-by: Yuchung Cheng <ycheng@google.com>
    Signed-off-by: Eric Dumazet <edumazet@google.com>
    Signed-off-by: Soheil Hassas Yeganeh <soheil@google.com>

cherusk:
To the former: cool, that would provide even more insights into CC in one swoop.

dtaht:
while I am here, iproute2 has a bug fixed recently:

commit d1f338b31858005b2550da05098d56ea4e8fd401
Author: Hangbin Liu <liuhangbin@gmail.com>
Date:   Thu Sep 22 16:40:28 2016 +0800

    misc/ss: tcp cwnd should be unsigned
    
    tcp->snd_cwd is a u32, but ss treats it like a signed int. This may
    results in negative bandwidth calculations.
    
    Signed-off-by: Hangbin Liu <liuhangbin@gmail.com>
    Acked-by: Phil Sutter <phil@nwl.cc>

dtaht:
And here:

commit 2f0f9aef94129643133363b4503468cdccc481cc
Author: Neal Cardwell <ncardwell@google.com>
Date:   Tue Sep 20 22:43:44 2016 -0400

    ss: output TCP BBR diag information
    
    Dump useful TCP BBR state information from a struct tcp_bbr_info that
    was grabbed using the inet_diag API.
    
    We tolerate info that is shorter or longer than expected, in case the
    kernel is older or newer than the ss binary. We simply print the
    minimum of what is expected from the kernel and what is provided from
    the kernel. We use the same trick as that used for struct tcp_info:
    when the info from the kernel is shorter than we hoped, we pad the end
    with zeroes, and don't print fields if they are zero.
    
    The BBR output looks like:
      bbr:(bw:1.2Mbps,mrtt:18.965,pacing_gain:2.88672,cwnd_gain:2.88672)
    
    The motivation here is to be consistent with DCTCP, which looks like:
      dctcp(ce_state:23,alpha:23,ab_ecn:23,ab_tot:23)
    
    Signed-off-by: Neal Cardwell <ncardwell@google.com>
    Signed-off-by: Yuchung Cheng <ycheng@google.com>
    Signed-off-by: Eric Dumazet <edumazet@google.com>
    Signed-off-by: Soheil Hassas Yeganeh <soheil@google.com>

cherusk:
So you're saying we're covering half the picture and most ss out there are off?

dtaht:
BBR is rather new. We aren't really grokking it yet. It is certainly interesting and I'd like to be able to understand it deeper. I'll patch netperf a bit more... but toke (and you) have the python skills to take a look at the existing sampling jitter better than I....

tohojo:
Dave Täht <notifications@github.com> writes:

> BBR is rather new. We aren't really grokking it yet. It is certainly
> interesting and I'd like to be able to understand it deeper. I'll
> patch netperf a bit more... but toke (and you) have the python skills
> to take a look at the existing sampling jitter better than I....

What sampling jitter are you talking about? The different durations of
each netperf sample?

Flent already uses negative values for -D

-Toke

dtaht:
I ran that netperf patch with the default flent values for netperf and saw several hundred ms worth of jitter.

tohojo:
Well, I guess netperf falls behind in its sampling? *shrugs* (but yeah, I'm seeing the same thing)

I have been meaning to get rid of the interpolated values and just plot the actual values output by netperf (at the time they appear); haven't got around to it yet. But Flent will capture it; `zgrep dur xxx.flent.gz` will show them.

tohojo:
As for the performance of ss sampling, I just pushed a couple of changes which will filter more in ss itself (cutting output in half), and also do duplicate suppression, so we don't spawn 50 identical ss_iterate for 50 flows, but just reuse the output from one. Quite feasible to enable --socket-stats with 50 flows now, at least on my machine.

Test, and maybe we can make --socket-stats the default if it works well enough?

cherusk:
@dtaht 
> BBR is rather new. We aren't really grokking it yet. It is certainly interesting and I'd like to be able to understand it deeper.

So do I(!), only thought it would confine itself to exporting only what with what other CCs are doing currently. We shall cover that.

@tohojo 
Good work! I've had a look at your recent changes and tested:

Up to **200** upload flows from my workstation . It was working for **~ 100 flows** once.

After that it kept leveling off at ~ 50 simultaneous flows with loosing the rest of the flows data.

Sometimes failing with:

> Stderr: 

> WARNING: Command produced no valid data.
> Runner class: NetperfDemoRunner
> Command: /usr/bin/netperf -P 0 -v 0 -D -0.20 -4  -H netperf-eu.bufferbloat.net -p 12865 -t TCP_STREAM -l 20 -f m   --   -H netperf-eu.bufferbloat.net 
> Return code: 255
> Stdout: establish control: are you sure there is a netserver listening on netperf-eu.bufferbloat.net at port 12865?
> establish_control could not establish the control connection from 0.0.0.0 port 0 address family AF_INET to netperf-eu.bufferbloat.net port 12865 address family AF_INET

That could be a problem on server side as well, right? Am I getting something wrong and those are the eliminated duplicates?

tohojo:
No, I think that is just netperf failing to start a flow because your link is so congested. That is normal(ish) :)

cherusk:
Might clamp it later into CORE to really stress test it, but see no value for that atm. Seems fine so far with what is there right now.

I've a further feature idea  ... shall I take it to the Mail-L.?

tohojo:
Sure, please do :)

cherusk:
Anyways, will start here, it will be one of you coming back anyways, I assume: 

What about not only spawning test flows but further tapping into already existing ones?

tohojo:
What does that mean?

-Toke

cherusk:
Not sure if it does add value:

One could tell flent which flows to follow already initialized by something else: any entity thinkable, e.g. an application. 

My notion somehow is that making thereby visible the effects of certain flow usage behaviours. Something very natural somebody would like to do with ss, actually. Though, that's something the tracers do by looking into the pkt flow (e.g. Wireshark with seq.-nums). Flent would do it with the eyes of CC. Might off of flents purpose or already represented in the RRUL tests.

tohojo:
Ah, you mean give Flent a PID, and it'll plot the TCP windows of all
TCP flows created by that PID? That might be useful :)

cherusk:
Essentially, that's the idea: the pid, or even pids. Thanks, for wrapping it up. :)

cherusk:
FYI: The gui is not depicting the gathered ss data reliably, at least not for all tests. There were no errs/warns. 

e.g. the case for:

cubic_westwood
cubic_bbr

The data is there:

> 
> TCP upload Cubic BE::tcp_cwnd     :        20.55        19.47                1819
>  TCP upload Cubic BE::tcp_rtt      :      1113.73      1126.60                1819
>  TCP upload Cubic BE::tcp_rtt_var  :        36.24        29.63                1819

tohojo:
Yeah, some of the tests lack the required include directive to get the
graphs... Will get around to fixing that at some point I guess ;)

PR #97, opened 2017-02-23
introduce tcp_cc_surveil test config

cherusk:
Add new test for tracing stream based flows of processes externally
started to flent.

Signed-off-by: Matthias Tafelmeier <matthias.tafelmeier@gmx.net>

### DIFF
--- a/flent/runners.py
+++ b/flent/runners.py
@@ -710,10 +710,10 @@ class NetperfDemoRunner(ProcessRunner):
                            'interval': args['interval'],
                            'length': args['length'],
                            'target': args['host'],
-                           'ip_version': args['ip_version']}
+                           'ip_version': args['ip_version'],
+                           'exclude_ports': [args['control_port']]}
 
-                self.add_child(SsRunner, [args['control_port']],
-                               ss_args, self.delay, None)
+                self.add_child(SsRunner, ss_args, self.delay, None)
 
         return "{binary} -P 0 -v 0 -D -{interval:.2f} -{ip_version} {marking} " \
             "-H {control_host} -p {control_port} -t {test} -l {length:d} " \
@@ -1022,8 +1022,15 @@ class SsRunner(ProcessRunner):
                  "LISTEN", "CLOSING"]
     ss_states_re = re.compile(r"|".join(ss_states))
 
-    def __init__(self, exclude_ports, command, *args, **kwargs):
-        self.exclude_ports = exclude_ports
+    def __init__(self, command, *args, **kwargs):
+        if 'exclude_ports' in kwargs.keys():
+            self.exclude_ports = kwargs['exclude_ports']
+            self.ext_run = False
+        else:
+            self._pid = kwargs['name'].split("::")[1]
+            self.exclude_ports = []
+            self.ext_run = True
+
         super(SsRunner, self).__init__(command, *args, **kwargs)
 
         dup_key = (command['host'], command['interval'], command['length'],
@@ -1060,7 +1067,7 @@ class SsRunner(ProcessRunner):
                            extra={'runner': self})
 
     def filter_np_parent(self, part):
-        sub_part = []
+        sub_parts_res = []
         sub_parts = self.ss_states_re.split(part)
         sub_parts = [sp for sp in sub_parts if sp.strip()
                      and not self.ss_header_re.search(sp)]
@@ -1078,39 +1085,79 @@ class SsRunner(ProcessRunner):
             if None is f_ports:
                 raise ParseException()
 
-            dst_p = int(f_ports.group('dst_p').split(":")[-1])
+            dst_shards = f_ports.group('dst_p').split(":")
+            # cure by checking ip_version
+            dst_addr = ""
+            if len(dst_shards) > 2:
+                dst_addr = ":".join(dst_shards[:-1])
+            else:
+                dst_addr = ".".join(dst_shards[:-1])
 
-            if self.par_pid == pid_str and dst_p not in self.exclude_ports:
-                sub_part.append(sp)
+            dst_p = dst_shards[-1]
 
-        if 1 != len(sub_part):
-            raise ParseException()
+            if self.filt_pid == pid_str and dst_p not in self.exclude_ports:
+                sub_parts_res.append((sp, (dst_addr, dst_p)))
 
-        return sub_part[0]
+        self.post_filter_pred(sub_parts_res)
 
-    def parse_part(self, part):
-        sub_part = self.filter_np_parent(part)
+        return sub_parts_res
 
-        timestamp = self.time_re.search(part)
+    def post_filter_pred(self, sub_parts):
+        if self.ext_run:
+            pass
+        else:
+            if 1 != len(sub_parts):
+                raise ParseException()
+
+    def form_sub_raw_vals(self, cwnd, rtt, rtt_var, flow_spec):
+        if self.ext_run:
+            sub_raw_val = {}
+            for k, val in zip(['cwnd', 'rtt', 'rtt_var'], [cwnd, rtt, rtt_var]):
+                key = "{1}#{2}::tcp_{0}".format(k, *flow_spec)
+                sub_raw_val[key] = val
+
+            return sub_raw_val
+        else:
+            return {'tcp_cwnd': cwnd,
+                    'tcp_rtt': rtt, 'tcp_rtt_var': rtt_var}
+
+    def parse_subpart(self, sub_part, **kwargs):
 
         cwnd = self.cwnd_re.search(sub_part)
         rtt = self.rtt_re.search(sub_part)
 
-        if None in [timestamp, cwnd, rtt]:
+        if None in [cwnd, rtt]:
             raise ParseException()
 
-        timestamp = float(timestamp.group('timestamp'))
         cwnd = int(cwnd.group('cwnd'))
         rtt, rtt_var = [float(x) for x in rtt.group('rtt').split('/')]
 
-        raw_values = {'t': timestamp, 'tcp_cwnd': cwnd,
-                      'tcp_rtt': rtt, 'tcp_rtt_var': rtt_var}
+        sub_raw_values = self.form_sub_raw_vals(cwnd, rtt, rtt_var,
+                                                kwargs['flow_spec'])
+
+        return sub_raw_values
+
+    def parse_part(self, part):
+        raw_values = {}
+        timestamp = self.time_re.search(part)
+        if None is timestamp:
+            raise ParseException()
+        timestamp = float(timestamp.group('timestamp'))
+
+        sub_parts = self.filter_np_parent(part)
+
+        for sp, flow_s in sub_parts:
+            sub_raw_values = self.parse_subpart(sp, flow_spec=flow_s)
+            for k, v in sub_raw_values.items():
+                raw_values[k] = v
+
+        raw_values['t'] = timestamp
         self._raw_values.append(raw_values)
 
         return raw_values
 
     def parse(self, output):
-        self.par_pid = str(self._parent.pid)
+        self.filt_pid = self.determine_pid()
         results = {}
         parts = output.split("\n---\n")
         for part in parts:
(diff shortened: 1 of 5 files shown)


cherusk:
**FYI - working on it from here.**

tohojo:
Wouldn't it be better to do this as a mixin (similar to qdisc stats) that can be added to any test?

cherusk:
My reasoning making it standalone was that we'll quite likely loose all the filtering optimizations in the ss script for that. And it appeared odd to me at first. But see the benefits ...

cherusk:
Well, Mr. @tohojo, it's working in that I'am not breaking the former SsRunner and the data points are properly gathered for the individual flows.

Open are:

- proper association of flows for plotting
- and plotting mechanisms
- config ok? - would have liked to integrate the tcp_stats.inc and the new .inc
- optimization: running only one ss iterator per local node and reuse full output

Kindly air your opinion when it suits you.

tohojo:
Couple of high-level comments:
- It should be possible to reuse the same runner instead of subclassing, by simply generalising it slightly.
- For the ss_script I think it would be simpler to just get rid of target and simple pass a filter; then have the Python code generate the full filter as appropriate.
- For the find_ss_iterate, just returning a dictionary will defer the call to find_binary (no need for a separate class method) - see find_netperf()
- Why can't you just reuse tcp_stats.inc?
- Use a longer test_parameter key (socket_stats_pids?) to make sure it is unambiguous
- Maybe add a socket_stats_ports also to filter out some ports?
- I don't particularly 'surveil' as a name; and the test you are adding is basically an empty test, so maybe just call it that? Could be useful for other things as well (qdisc stats, wifi stats, etc).

cherusk:
> - It should be possible to reuse the same runner [...]
Tried that first. Got ugly, especially for the initialization part. Could get rid of some conditionals as well.
> For the ss_script I think it would be simpler to just get rid of target and simple pass a filter [...]
Would embrace to, but what to filter for? Filterting for PID is not supported by ss up to now. And the assumption is, I don't know anything of the flows outline of the targeted process.
> For the find_ss_iterate
Would make sense without derivative, but the behavior for the surveillance ss run differs, so I'd need a override anyways, I assume.
> Why can't you just reuse tcp_stats.inc?
Thought because of the test series plot naming. Am not into your plotting conf yet.
>Use a longer test_parameter key [...]
Ok ...
> Maybe add a socket_stats_ports also to filter out some ports?
From performance yes, from semantics rather not. Adore the notion of being in position to ignore the flow outline and in my mind that's the only sensible use case.
> [...] particularly 'surveil' as a name [...]
What's wrong with surveil like surveillance? Am indifferent, as such.
> Could be useful for other things as well
Aeh, yea, I have already an .inc for that then, right! See?

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

>  * It should be possible to reuse the same runner [...]
>  Tried that first. Got ugly, especially for the initialization part.
>  Could get rid of some conditionals as well.

Why? You just need to pass the pid to the constructor, and then only get
it from the parent runner if it's not set already? What else is different?

-Toke

cherusk:
> What else is different? 

The __init__ API itself, for example?! Different code paths using it!?

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

>  What else is different?
>
> The init API itself, for example?! Different code paths using it!?

What's wrong with something like this?

def __init__(pid=None...):
  self.parse_pid = pid

def parse(...):
  if self.parse_pid is None and self._parent:
    self.parse_pid = self._parent.pid

cherusk:
Know what you were getting at. Thought of this discrminator. 

Aside from remodeling what inheritance does, I'd have to stuff the exclude_ports into kwargs.

Well, not more as such for the __init.

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

> Know what you were getting at. Thought of this discrminator.
>
> Aside from remodeling what inheritance does, I'd have to stuff the
> exclude_ports into kwargs.

Sure, do that. It may be that it was a mistake to have add_child pass
positional arguments instead of keyword arguments; feel free to fix that
if it makes things easier.

And conceptually, the SsRunner parses ss output filtering on pid and
optionally excluding some ports. That should be general enough to cover
both the existing use case and the "follow and external pid" case,
without needing a subclass. Only issue is what to do with the data when
more than one socket is found. That depends on how we treat the data; do
you envision that a user selects a pid, and then each flow generated by
that pid gets its own graph, keyed on port number?

-Toke

cherusk:
> [...] do you envision that a user selects a pid, and then each flow generated by that pid gets its own graph, keyed on port number?

Yes, this is exactly the way I've modeled it in code by now. Appears quite clean.

cherusk:
E.g.
$ ./run-flent -v -l 20 -4 tcp_cc_surveil localhost --test-parameter socket_stats_pids=3317

leads to 

                                                             avg       median          # data pts
 sock_stats::3317::tcp_cwnd_193.175.141.61_143    :        10.00        10.00                 103
 sock_stats::3317::tcp_cwnd_194.25.134.114_993    :        10.00        10.00                 103
 sock_stats::3317::tcp_cwnd_212.227.17.170_993    :        10.00        10.00                 103
 sock_stats::3317::tcp_rtt_193.175.141.61_143     :        39.24        39.24                 103
[...]

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

> E.g.
> $ ./run-flent -v -l 20 -4 tcp_cc_surveil localhost --test-parameter socket_stats_pids=3317
>
> leads to
>
>                                                          avg       median          # data pts
>
> sock_stats::3317::tcp_cwnd_193.175.141.61_143 : 10.00 10.00 103

Hmm, if this was sock_stats::3317::193.175.141.61_143::tcp_cwnd instead,
then it would work with the globbing in the existing plots. And we
should really have a separate entry in raw_values for each socket...

-Toke

cherusk:
Cool. Weren't the test_parameter, the includes would be mergeable then.

cherusk:
Hold on, would have to know flow in advance during conf for connecting data to null_runnner. 

DATA_SETS['sock_stats::%s::tcp_cwnd' % pid]

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

> Hold on, would have to know flow in advance during conf for connecting data to null_runnner.
>
> DATA_SETS['sock_stats::%s::tcp_cwnd' % pid]

Yeah, I'm working on that. This part is a bit of a wart, and I'm
reworking the plotting code so it won't be necessary...

-Toke

cherusk:
Getting stuck here for SsRunner:

  File "./run-flent", line 33, in <module>
    sys.exit(run_flent())
  File "/home/matthias/Projects/sync/cc_qdisc/flent/flent/__init__.py", line 56, in run_flent
    b.run()
  File "/home/matthias/Projects/sync/cc_qdisc/flent/flent/batch.py", line 598, in run
    return self.run_test(self.settings, self.settings.DATA_DIR, True)
  File "/home/matthias/Projects/sync/cc_qdisc/flent/flent/batch.py", line 497, in run_test
    res = self.agg.postprocess(self.agg.aggregate(res))
  File "/home/matthias/Projects/sync/cc_qdisc/flent/flent/aggregators.py", line 258, in aggregate
    measurements, metadata, raw_values = self.collect()
  File "/home/matthias/Projects/sync/cc_qdisc/flent/flent/aggregators.py", line 138, in collect
    self.threads[n] = i['runner'](name=n, settings=self.settings, **i)
TypeError: __init__() takes exactly 2 arguments (1 given)

External startup quite differs from sub childing.

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

> Getting stuck here for SsRunner:
>
> Traceback (most recent call last):
> File "./run-flent", line 33, in 
> sys.exit(run_flent())
> File "/home/matthias/Projects/sync/cc_qdisc/flent/flent/init.py", line 56, in run_flent
> b.run()
> File "/home/matthias/Projects/sync/cc_qdisc/flent/flent/batch.py", line 598, in run
> return self.run_test(self.settings, self.settings.DATA_DIR, True)
> File "/home/matthias/Projects/sync/cc_qdisc/flent/flent/batch.py", line 497, in run_test
> res = self.agg.postprocess(self.agg.aggregate(res))
> File "/home/matthias/Projects/sync/cc_qdisc/flent/flent/aggregators.py", line 258, in aggregate
> measurements, metadata, raw_values = self.collect()
> File "/home/matthias/Projects/sync/cc_qdisc/flent/flent/aggregators.py", line 138, in collect
> self.threads[n] = i['runner'](name=n, settings=self.settings, **i)

**i here is basically everything that you put into the dictionary in
your test definition. So make sure you either have everything
(exclude_ports and pids) in the test definition, or fill in defaults
find_ss_runner (like find_netperf does).

-Toke

cherusk:
No, no, I've already done that. I've traced **i 

WARNING: {u'finish_event': <threading._Event object at 0x7f27bec1b3d0>, u'idx': 0, u'runner': <class 'flent.runners.SsRunner'>, u'delay': 0, u'command': {u'ip_version': 4, u'host': u'localhost', u'interval': 0.2, u'length': 20}, u'kill_event': None, u'units': u'misc', u'start_event': None, u'remote_host': None}

The thing is the differing __init__ api, I guess:

For SsRunner:
def __init__(self, cmdargs, *args, **kwargs):
For SssurveilRunner (working)
def __init__(self, *args, **kwargs):

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

> No, no, I've already done that. I've traced **i
>
> WARNING: {u'finish_event': <threading._Event object at 0x7f27bec1b3d0>, u'idx': 0, u'runner': <class 'flent.runners.SsRunner'>,
> u'delay': 0, u'command': {u'ip_version': 4, u'host': u'localhost', u'interval': 0.2, u'length': 20}, u'kill_event': None, u'units': u'misc',
> u'start_event': None, u'remote_host': None}
>
> The thing is the differing init api, I guess:
>
> For SsRunner:
> def init(self, cmdargs, *args, **kwargs):

Ah, yes, that is a bug. Pushed a fix :)

-Toke

cherusk:
Ok, thanks, looks fine now:

Only the depiction left ...
First gui run meant - 

> ''Settings' object has no attribute 'DEFAULTS''

cherusk:
FYI:

That was missing in my conf:

DEFAULTS={'PLOT': 'tcp_cwnd'}
PLOTS=o()

No errs then, but data depiction was blank. :D

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

> FYI:
>
> That was missing in my conf:
>
> DEFAULTS={'PLOT': 'tcp_cwnd'}
> PLOTS=o()
>
> No errs then, but data depiction was blank. :D

Yeah, not sure we can fix that properly until I finish up the plotting
rework...

-Toke

cherusk:
No probs, could you merge that in before, please? Otherwise falling behind master again ...

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

> No probs, could you merge that in before, please? Otherwise falling
> behind master again ...

Yeah, I'll merge it as soon as it's done. I can prioritise getting the
parts done you'll need to get plots working, and push that to a separate
branch for you to work from. No promises as to when I'll get those parts
done, though; it's a somewhat big rework of the plotting code.

-Toke

cherusk:
> I can prioritise getting the
> parts done you'll need to get plots working, and push that to a separate
> branch for you to work from. 
Yep, a upstream branch would help.

> No promises as to when I'll get those parts
> done, though; it's a somewhat big rework of the plotting code.
No pressure from my end! The first PR was prior! Guess cannot help you,
since comm. effort might higher than rework.

-- 
Besten Gruß

Matthias Tafelmeier

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

>> I can prioritise getting the
>> parts done you'll need to get plots working, and push that to a separate
>> branch for you to work from. 
> Yep, a upstream branch would help.
>
>> No promises as to when I'll get those parts
>> done, though; it's a somewhat big rework of the plotting code.
> No pressure from my end! The first PR was prior! Guess cannot help you,
> since comm. effort might higher than rework.

In this instance, yeah. I'll ping you when I've pushed something :)

-Toke

cherusk:
Sorry for poking, found that in my todos. Are you working on it?

tohojo:
Matthias Tafelmeier <notifications@github.com> writes:

> Sorry for poking, found that in my todos. Are you working on it?

Haven't had time to finish it; sorry about that. Will see if I can find
the time soonish :)

-Toke

tohojo:
Okay, pushed the raw-plotting branch; sorry for taking so long to get the rest done.

In that branch you can specify a raw_key in plot configurations to plot different values in the same data series. So you can have a data series for each socket, and you don't need the null runners.

This still only works for timeseries plots, and I may have to change things around as I progress on the other types of plots. So consider the code in that branch somewhat unstable, but feel free to build on it :)

PR #110, opened 2017-05-01
tcp_cc_surveil (3)

cherusk:
Repeat>
Ok, works fairly enough. A minor one, did not find where the generic label is altered, so for every flow that one is depicted. ^^

### DIFF
--- a/flent/runners.py
+++ b/flent/runners.py
@@ -710,10 +710,10 @@ class NetperfDemoRunner(ProcessRunner):
                            'interval': args['interval'],
                            'length': args['length'],
                            'target': args['host'],
-                           'ip_version': args['ip_version']}
+                           'ip_version': args['ip_version'],
+                           'exclude_ports': [args['control_port']]}
 
-                self.add_child(SsRunner, [args['control_port']],
-                               ss_args, self.delay, None)
+                self.add_child(SsRunner, ss_args, self.delay, None)
 
         return "{binary} -P 0 -v 0 -D -{interval:.2f} -{ip_version} {marking} " \
             "-H {control_host} -p {control_port} -t {test} -l {length:d} " \
@@ -1022,8 +1022,16 @@ class SsRunner(ProcessRunner):
                  "LISTEN", "CLOSING"]
     ss_states_re = re.compile(r"|".join(ss_states))
 
-    def __init__(self, exclude_ports, command, *args, **kwargs):
-        self.exclude_ports = exclude_ports
+    def __init__(self, command, *args, **kwargs):
+        if 'exclude_ports' in command:
+            self.exclude_ports = command['exclude_ports']
+            self.ext_run = False
+            del command['exclude_ports']
+        else:
+            self._pid = kwargs['name'].split("::")[1]
+            self.exclude_ports = []
+            self.ext_run = True
+
         super(SsRunner, self).__init__(command, *args, **kwargs)
 
         dup_key = (command['host'], command['interval'], command['length'],
@@ -1062,7 +1070,7 @@ class SsRunner(ProcessRunner):
                            extra={'runner': self})
 
     def filter_np_parent(self, part):
-        sub_part = []
+        sub_parts_res = []
         sub_parts = self.ss_states_re.split(part)
         sub_parts = [sp for sp in sub_parts if sp.strip()
                      and not self.ss_header_re.search(sp)]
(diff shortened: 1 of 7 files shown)


flent/runners.py
```diff
@@ -1006,6 +1006,13 @@ class SsRunner(ProcessRunner):
     time_re = re.compile(r"^Time: (?P<timestamp>\d+\.\d+)", re.MULTILINE)
     cwnd_re = re.compile(r"cwnd:(?P<cwnd>\d+)", re.MULTILINE)
     rtt_re = re.compile(r"rtt:(?P<rtt>\d+\.\d+/\d+\.\d+)", re.MULTILINE)
+
+    bbr_re_str = "bbr:\(bw:(?P<bw>\d+\.\d+)(?P<bw_scale>.bps),"\
+                 "mrtt:(?P<mrtt>\d+\.\d+),"\
+                 "pacing_gain:(?P<pacing_gain>\d+\.\d+),"\
+                 "cwnd_gain:(?P<cwnd_gain>\d+)\)"
```
tohojo: You are repeating these keys in three different places in the code. Use match.groupdict() instead to just pull them out of the regex match...

flent/scripts/ss_iterate.sh
```diff
@@ -18,31 +18,39 @@ count=10
 interval=0.1
 host=localhost
 filter=""
+surveil=""
 
 usage()
 {
-    echo "$0 -c count -I interval -H host -t target"
+    echo "$0 -c count -I interval -H host -t target -f filter"
```
tohojo: Hmm, instead of adding yet another arg here, maybe just get rid of -t and move the "dst $TARGET" into the Python code as part of the filter?

flent/runners.py
```diff
@@ -1080,39 +1095,108 @@ def filter_np_parent(self, part):
...
+            pass
+        else:
+            if 1 != len(sub_parts):
+                raise ParseException()
+
+    def form_sub_raw_vals(self, cwnd, rtt, rtt_var, bbr, flow_spec):
+        bbr_keys = ['bbr_bw', 'bbr_mrtt', 'bbr_pacing_gain', 'bbr_cwnd_gain']
+        if self.ext_run:
+            sub_raw_val = {}
+            for k, val in zip(['cwnd', 'rtt', 'rtt_var'], [cwnd, rtt, rtt_var]):
```
tohojo: You can probably get rid of the repetition here in a similar way (named groups in regexes + groupdict())

flent/tests/tcp_cc_surveil.inc
```diff
@@ -0,0 +1,46 @@
...
+PLOTS['tcp_cwnd'] = {'description': 'TCP window statistics',
+                     'type': 'timeseries',
+                     'axis_labels': ['Packets'],
+                     'series': [
+                         {
+                          'data': glob("sock_stats::**"),
+                          'label': 'TCP window size',
+                          'raw_key' : glob("**::tcp_cwnd")
+                      },
+                     ]}
```
tohojo: It would be good if we could re-use the plot definitions in tcp_stats.inc - may need a bit of restructuring, though...

tohojo:
Okay, the review comments got a bit mangled (I really do not like the github interface for this :/). But the top comment was supposed to be for the "introduce globbing to raw_keys" commit, and the comments on specific code lines were meant to go together...

cherusk:
On 05/10/2017 04:10 PM, Toke Høiland-Jørgensen wrote:
>
> *@tohojo* commented on this pull request.
>
> NAK on this whole commit. Introducing another level of globbing is not
> the right way to do this;
>
Well, it's quite flexible, especially for the surveillance case at hand:

Could filter on flow plots by extending the suffix, that was part of the
rational.

E.g.:
Instead of ...
[...]
   'series': [
                         {
                          'data': glob("sock_stats::**"),
                          'label': 'TCP window size',
         >>>          'raw_key' : glob("**::tcp_cwnd")
                      },

[...]
   'series': [
                         {
                          'data': glob("sock_stats::**"),
                          'label': 'TCP window size',
         >>>          'raw_key' : glob("*#433*::tcp_cwnd")
                      },

Selecting all those *dest-pt* flows for plotting and so forth.

> instead, create separate series that each have one set of raw_values.
>
Hmmm, but that is the end what the globbing mechanism is doing, as I
understood it's working. What's cleaner about it by doing it seperately?

> ------------------------------------------------------------------------
>
> In flent/runners.py
> <https://github.com/tohojo/flent/pull/110#discussion_r115749917>:
>
> > @@ -1006,6 +1006,13 @@ class SsRunner(ProcessRunner):
>      time_re = re.compile(r"^Time: (?P<timestamp>\d+\.\d+)", re.MULTILINE)
>      cwnd_re = re.compile(r"cwnd:(?P<cwnd>\d+)", re.MULTILINE)
>      rtt_re = re.compile(r"rtt:(?P<rtt>\d+\.\d+/\d+\.\d+)", re.MULTILINE)
> +
> +    bbr_re_str = "bbr:\(bw:(?P<bw>\d+\.\d+)(?P<bw_scale>.bps),"\
> +                 "mrtt:(?P<mrtt>\d+\.\d+),"\
> +                 "pacing_gain:(?P<pacing_gain>\d+\.\d+),"\
> +                 "cwnd_gain:(?P<cwnd_gain>\d+)\)"
>
> You are repeating these keys in three different places in the code.
> Use match.groupdict() instead to just pull them out of the regex match...
>
ACK, that's opening a maintenace hell. Will rectifiy!


-- 
Besten Gruß

Matthias Tafelmeier

flent/tests/tcp_cc_surveil.inc
cherusk: On top, I tend to decist from mixing active code with configuration as done in tcp_stats.inc.

cherusk:
This has become stale. Is there any way forward or has it went unfathomable?

tohojo:
You should be able to rebase. There's still an issue with how to store and display the data per pid. I think the solution is to create a multi-level dictionary (i.e. raw_values[pid]['tcp_rtt'] will have the RTT val). This is how the wifi stats data is stored. Still haven't gotten around to implementing support for plotting that in raw_values, but I'll have to do that anyway for the wifi stats...

cherusk:
On 06/20/2017 12:01 PM, Toke Høiland-Jørgensen wrote:
>
> You should be able to rebase.
>
Theoretically ... this has diverged too wild by now.


-- 
BR

Matthias Tafelmeier

PR #125, opened 2017-12-12
Use = with all long form flags with values.

peteheist:
For some reason, the Go pflag packages (both of them) require long form flags with values to have equals signs. This works for me, but I'm not sure of and didn't test the local_bind stuff in particular, so by all means fix or replace any of it if there's a better way. :)

### DIFF
--- a/flent/runners.py
+++ b/flent/runners.py
@@ -1477,9 +1477,9 @@ class IrttRunner(ProcessRunner):
                     '--timeouts=200ms,300ms,400ms']
 
             if self.local_bind:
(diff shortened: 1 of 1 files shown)


tohojo:
Ah right, forgot about that. Only comment on this is that the single-element extend calls should be changed to append without the wrapping in a list.

Also, please add a signed-off-by line to the commit message :)

peteheist:
Cool, hope that works...

tohojo:
Just fixed this manually and merged to master...

PR #126, opened 2018-01-05
Let run-flent to run gui and mention in README

mlouielu:
To make `run-flent` script can run gui directly from source code.


mlouielu:
After reading `gui.rst`, this should be polish to only documentation enhance in `quick-start.rst`.

PR #141, opened 2018-07-12
Fix wrong equation in the E-Model

yehcj:
The equation of the ```Idd``` is wrong which leads to wrong MOS.


tohojo:
Good catch, thanks!

Do you have any idea how big the errors in MOS values from this bug are?

yehcj:
The MOS will downgrade to 1 when delay greater than 100ms. 

When loss = 0,
- delay = 100, MOS = 4.X
- delay = 101, MOS = 1.X

I think this is a severe problem.

tohojo:
Heh, yeah, that's bad... :/

Thanks for taking the time to report this!

PR #164, opened 2019-06-28
Fix executed command when using remote-host option

remip2:
When using remote-host option to run stream generators on another hosts, the commands are executed on the same host flent is running. The commands are prefixed with ssh, but the forked process run self.args which is defined just before the update.


tohojo:
Thanks for the report, and fix!

There were a few style issues with the pull request (missing DCO, no commit
message, you forgot to move the metadata saving of the command), so I just
committed this fix instead of merging the pull request.

tohojo:
Fixed in 276abd7e28d264e76d8b739f7e3867710cfbedbc

PR #168, opened 2019-07-20
Fix pacing and delivery rate calculations and add plot for tcp_nup

heistp:
Jon and I noticed this with pacing and delivery rate, and this plot might be useful for others as well. :)


tohojo:
Heh, division, multiplication, same same, right? Oops. Thanks for catching (and
fixing) this :)

PR #188, opened 2019-11-02
metadata: Add parsing of WiFi status information

emys-alb:
Created a function called get_wifi_data() that gathers information about the wifi status at one dictionary.

Output:
> {'addr': '50:b7:c3:c9:1c:8c',
 'channel': {'band': 2462, 'center1': 2462, 'number': 11, 'width': 20},
 'ssid': 'VIVO-81A0',
 'txpower': 15.0,
 'type': 'managed'}

### DIFF
--- a/flent/metadata.py
+++ b/flent/metadata.py
@@ -138,7 +138,8 @@ def record_metadata(results, extended, hostnames):
     if extended:
         m['IP_ADDRS'] = get_ip_addrs()
         m['GATEWAYS'] = get_gateways()
-
+        m['WIFI_DATA'] = get_wifi_data()
+        
     m['REMOTE_METADATA'] = {}
 
     for h in hostnames:
@@ -166,8 +167,9 @@ def record_metadata(results, extended, hostnames):
         if extended:
             m['REMOTE_METADATA'][h]['IP_ADDRS'] = get_ip_addrs()
             m['REMOTE_METADATA'][h]['GATEWAYS'] = get_gateways()
+            m['REMOTE_METADATA'][h]['WIFI_DATA'] = get_wifi_data()
 
-
+            
 def record_postrun_metadata(results, extended, hostnames):
     m = results.meta()
     get_command_output.set_hostname(None)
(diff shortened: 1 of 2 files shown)


flent/metadata.py
```diff
@@ -514,3 +514,27 @@ def get_module_versions():
             module_versions[m] = v
 
     return module_versions
+
+def get_wifi_data(iface):
```
tohojo: You'll also need to add appropriate calls to this function in record_metadata(); I'd suggest putting them along with the calls to get_ip_addrs() and get_gateways()

flent/metadata.py
```diff
@@ -514,3 +514,27 @@ def get_module_versions():
...
+    wifi_data = {}
+    unwanted_keys = ["Interface", "ifindex", "wdev", "wiphy"]
+    output = get_command_output("iw dev %s info" % iface)
+    if output is not None:
+        for line in output.splitlines():
+            parts = line.split()
```
tohojo: This needs to discard empty lines first

flent/metadata.py
```diff
@@ -514,3 +514,27 @@ def get_module_versions():
...
+    wifi_data = {}
+    unwanted_keys = ["Interface", "ifindex", "wdev", "wiphy"]
+    output = get_command_output("iw dev %s info" % iface)
+    if output is not None:
+        for line in output.splitlines():
+            parts = line.split() 
+
+            if parts[0]  in unwanted_keys:
+                continue
+            k,v = parts[0], parts[1]
```
tohojo: Since you're doing the split here, better to use the 'k' and 'v' variables further down instead of parts[0] and parts[1]

flent/metadata.py
```diff
@@ -514,3 +514,27 @@ def get_module_versions():
...
+            if parts[0] == 'txpower':
+               v = float(parts[1])
+            if parts[0] == 'channel':
```
tohojo: Please add a comment with an example of the line you are trying to parse with this

flent/metadata.py
```diff
@@ -514,3 +514,27 @@ def get_module_versions():
...
+                v = {}
+                v['number'] =int(parts[1])
+                v['band'] = int(parts[2].strip("("));
+                v['width'] = int(parts[5])
+                v['center1'] = int(parts[8])
+                
+            wifi_data[k] = v
+
+    return wifi_data
```
tohojo: Missing newline at the end of the file

flent/metadata.py
tohojo: Also, you'll need to handle unwanted lines. On my system, the output looks like this:

```
Interface wlan0
	ifindex 3
	wdev 0x1
	addr 98:de:ad:ca:fe:85
	ssid myssid
	type managed
	wiphy 0
	channel 36 (5180 MHz), width: 80 MHz, center1: 5210 MHz
	txpower 22.00 dBm
	multicast TXQ:
		qsz-byt	qsz-pkt	flows	drops	marks	overlmt	hashcol	tx-bytes	tx-packets
		0	0	0	0	0	0	0	0		0
```
Your parser will choke on the bit at the bottom; I think it's probably fine if you just abort parsing when you get to the "multicast TXQ" line...

flent/metadata.py
emys-alb: I didn't get it

flent/metadata.py
emys-alb: like 'txpower: 15.0'?

flent/metadata.py
tohojo: You added a new function. But it doesn't do anything until you add some code that calls that function and does something with the result. Look at the how get_ip_addrs() is being called in record_metadata() for an example...

flent/metadata.py
tohojo: No, I meant the 'channel' line...

PR #192, opened 2019-11-26
metadata: Add parsing of WiFi status information

emys-alb:
Created a function called get_wifi_data() that gathers information about the wifi status at one dictionary.

Output:

    {'addr': '50:b7:c3:c9:1c:8c',
    'channel': {'band': 2462, 'center1': 2462, 'number': 11, 'width': 20},
    'ssid': 'VIVO-81A0',
    'txpower': 15.0,
    'type': 'managed'}

### DIFF
--- a/flent/metadata.py
+++ b/flent/metadata.py
@@ -138,6 +138,7 @@ def record_metadata(results, extended, hostnames):
     if extended:
         m['IP_ADDRS'] = get_ip_addrs()
         m['GATEWAYS'] = get_gateways()
+        m['WIFI_DATA'] =get_wifi_data()
 
     m['REMOTE_METADATA'] = {}
 
@@ -166,6 +167,7 @@ def record_metadata(results, extended, hostnames):
         if extended:
             m['REMOTE_METADATA'][h]['IP_ADDRS'] = get_ip_addrs()
             m['REMOTE_METADATA'][h]['GATEWAYS'] = get_gateways()
+            m['REMOTE_METADATA'][h]['WIFI_DATA'] = get_wifi_data()
 
 
 def record_postrun_metadata(results, extended, hostnames):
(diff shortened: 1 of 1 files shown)


flent/metadata.py
```diff
@@ -514,3 +516,39 @@ def get_module_versions():
...
+
+                if k  in unwanted_keys:
+                  continue
+
+                if k == 'txpower':
+                   v = float(parts[1])
+
+                if k == 'channel':
+                    # This condition will return a dict with all the values of the channel
+		            # Output will be {'addr':..., channel': {'band': 2462, 'center1': 2462, 'number': 11, 'width': 20}, 'ssid':...}
```
tohojo: The indentation is off on the second comment line. Also, please add an example of the input (i.e., the line you're trying to parse) along with the output.

flent/metadata.py
```diff
@@ -514,3 +516,39 @@ def get_module_versions():
...
+                    v = {}
+                    v['number'] =int(parts[1])
+                    v['band'] = int(parts[2].strip("("));
+                    v['width'] = int(parts[5])
+                    v['center1'] = int(parts[8])
+
+                if k == "multicast TXQ":
```
tohojo: Please add a short comment explaining this (could just be, e.g., "No interesting output after this")

flent/metadata.py
```diff
@@ -514,3 +516,39 @@ def get_module_versions():
...
+                    break
+
+            wifi_data[k] = v
```
tohojo: This should be inside the if statement.

Alternatively, you could save one level of indentation by reversing the if statement above and turning it into a "continue". I.e., replace

```
if len(line.strip()) >= 1:
   # lots of code here
```
with
```
if not line.strip():
  continue
# code here (unindented)
```

Up to you which you prefer :)

flent/metadata.py
```diff
@@ -138,6 +138,7 @@ def record_metadata(results, extended, hostnames):
     if extended:
         m['IP_ADDRS'] = get_ip_addrs()
         m['GATEWAYS'] = get_gateways()
+        m['WIFI_DATA'] =get_wifi_data()
```
tohojo: Missing space after =

emys-alb:
The second one has some changes in the code and I wasn't sure if --amend
would amend everything or just the commit message.

emys-alb:
And thank you very much!!

tohojo:
git --amend will also add any new changes to the existing commit, not just change the commit message. It's a way to keep the history "clean" (having one logical commit) while making changes.

tohojo:
Huh, I guess Github isn't smart enough to figure out that that commit was a manual merge of the PR, so it shows up as "closed" instead of "merged". But don't worry, the actual changes are merged :)

Also, I fixed up the author name on your commit to use your full name; that is good practice, so you may want to update your git config to do this by default. See https://help.github.com/en/github/using-git/setting-your-username-in-git

PR #200, opened 2020-03-09
tests: Add tcp_delivery_with_rtt and upload_box plots.

heistp:


flent/tests/tcp_nup.conf
```diff
@@ -56,6 +56,14 @@ PLOTS = o([
...
                   'color': 'black',
                   'linewidth': 2}]}),
+    ('upload_box',
+     {'description': 'Upload box plot',
+      'type': 'box',
+      'series': [{'data': glob('TCP upload::*'),
+                  'label': 'BE'},
+                 {'data': 'TCP upload avg',
+                  'label': 'Avg',
+                  'cutoff': (DELAY,-DELAY)}]}),
```
tohojo: Shouldn't cutoff be consistent between the average and the individual flows?

heistp:
Should be fixed. I admit that I haven't actually tested with a cutoff, but the plot works...

tohojo:
Pete Heist <notifications@github.com> writes:

> Should be fixed. I admit that I haven't actually tested with a cutoff,
> but the plot works...

The plot will work in any case (as in it'll display something); it's
just a matter of which subset of data will be part of it :)

PR #205, opened 2020-05-08
Add TCP delivery rate plots to rrul_be

heistp:
Hi, just adding a few plots we're using with rrul_be...


tohojo:
Hmm, the tests fail because this new plot type doesn't have any output when plotted from the existing data files in unittests/test_data.

To fix this, could you please:
- Add the new plot name to PLOTS_MAY_FAIL in unittests/test_plotters.py
- Add a new data file in unittests/test_data with one of your data files using this plot?

heistp:
So that's there. Might have just needed an rrul_be test with socket stats, not sure.

tohojo:
Nah, it tries to plot all the plots with all the test data files. Probably ought to make it a bit smarter, but for now could you please just add all the new plot names to PLOTS_MAY_FAIL? And also add s-o-b tags to the two last commits (or just squash everything)?

heistp:
Ok, let's see if I finally got it right.

Issue #238, opened 2021-09-27
irtt fails with IPv6 (wrong format)


j-breyer:
Testing flent with IPv6 Addresses yields a similar error to [Issue #214](https://github.com/tohojo/flent/issues/214), despite installing the irtt tool. (IPv4 works fine with irtt, already tested.)

Running in `-v` to verbose showed me, that the format of IPv6 addresses is false for `irtt`:
Part of the output:
`which: Found irtt executable at /usr/bin/irtt`
`UDP RTT test: Cannot use irtt runner (Irtt connection check failed: Error: address fe80::200:aff:fe00:601%h1-eth0:0: too many colons in address`
`). Using netperf UDP_RR`
`Ping (ms) UDP EF: Adding child NetperfDemoRunner`

flent is run with `--local-bind` and `-H` in my case, for address handover of client and server.

Scrolling a bit through the [irtt docs](https://htmlpreview.github.io/?https://github.com/heistp/irtt/blob/master/doc/irtt-client.html#host-formats), I found that the host format for IPv6 addresses should be in brackets [ ]

Handing over the `--local-bind` in bracket format as a quick test solves the irtt issue with that address, but of course results in the address resolution to fail

Any help or quick fix is much appreciated!

tohojo:
Could you please include the full command you're running flent with, and the full log output? I'm not quite sure I understand what you're trying to do and what goes wrong from your description...

j-breyer:
I'm running the flent tests in a mininet with IPv6 enabled (ping is working).
On server side:
`sudo netserver -6 -L fe80::200:aff:fe00:e02%h2-eth0 -p 12865 -d`
to start the netserver, then
`sudo irtt server`
for the irtt server

On client side:
`sudo flent rrul -6 -p all_scaled --local-bind=fe80::200:aff:fe00:601%h1-eth0 -l 60 -H fe80::200:aff:fe00:e02 -t ipv6 -o ipv6_dscp.pdf -v`
fails to achieve the UDP RTT measurement, but completes the TCP and ICMP6 part (full verbose output below).

Running the same command with IPv4 addresses of the hosts (netserver enabled respectively):
`sudo flent rrul -p all_scaled --local-bind=10.0.6.1 -l 60 -H 10.0.14.2 -t ipv4 -o ipv4_dscp.pdf -v`
works just fine.

The difference in output comes in these lines:
```
Ping (ms) UDP EF: Adding child IrttRunner
[...]

PR #239, opened 2021-09-29
runners: Bracket IPv6 addresses before passing to irtt

heistp:
irtt currently only accepts IPv6 addresses surrounded by brackets.
If the host contains an IPv6 address literal (as identified by
ipaddress.IPv6Address), surround the host with brackets before
passing to irtt.

This should address #238.

### DIFF
--- a/flent/runners.py
+++ b/flent/runners.py
@@ -1559,7 +1559,7 @@ class IrttRunner(ProcessRunner):
     def __init__(self, host, length, interval=None, ip_version=None,
                  local_bind=None, marking=None, multi_results=False,
                  sample_freq=0, data_size=None, **kwargs):
-        self.host = normalise_host(host)
+        self.host = normalise_host(host, bracket_v6=True)
         self.interval = interval
         self.length = length
         self.ip_version = ip_version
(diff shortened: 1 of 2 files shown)


flent/runners.py
```diff
@@ -1682,7 +1683,13 @@ def check(self):
             if self.ip_version is not None:
                 args.append("-{}".format(self.ip_version))
 
-            args.append(self.host)
+            try:
+                ipaddress.IPv6Address(self.host)
+                host = "[{}]".format(self.host)
+            except ValueError:
+                host = self.host
```
tohojo, line 1690: Using the ipaddress module for parsing is fine, but let's move this into the normalise_host() utility function. Simply adding a second argument 'bracket_v6' which defaults to 'False' to that should do the trick :)

tohojo (changes requested):

flent/runners.py
heistp: Ah, I didn't notice that, that's a much better solution. :)

heistp: It just occurred to me that when adding optional parameters, it might be good practice to specify them by name in calls to those functions, so I updated that. I don't know what normal python practice is. I guess as long as the parameter positions don't change it's not critical, but this might also make it more apparent what you're doing in the call itself.

tohojo (approved):
LGTM, but could you please squash the changes into a single commit? :)

tohojo:
Awesome - thanks! :)

PR #243, opened 2021-10-19
Added --test-payload option

hrishikeshathalye:
Added the --test-payload option for sending a custom payload by using netperf's buffer prefill option

### DIFF
--- a/flent/runners.py
+++ b/flent/runners.py
@@ -983,6 +983,7 @@ class NetperfDemoRunner(ProcessRunner):
         args.setdefault('send_size',
                         self.settings.SEND_SIZE[0]
                         if self.settings.SEND_SIZE else "")
+        args.setdefault('test_payload', self.settings.TEST_PAYLOAD)
 
         if self.settings.SWAP_UPDOWN:
             if self.test == 'TCP_STREAM':
@@ -1022,12 +1023,19 @@ class NetperfDemoRunner(ProcessRunner):
                 netperf['-e'] = True
 
             try:
-                # Sanity check; is /dev/urandom readable? If so, use it to
+                # If --test-payload option is specified, use data from that file
+                # else use the default value /dev/urandom. 
+                fill_file = args['test_payload']
+                # Sanity check; is /dev/urandom or the custom file readable? If so, use it to
                 # pre-fill netperf's buffers
-                self.run_simple(['dd', 'if=/dev/urandom', 'of=/dev/null', 'bs=1', 'count=1'], errmsg="Err")
-                netperf['buffer'] = '-F /dev/urandom'
+                self.run_simple(['dd', 'if='+fill_file, 'of=/dev/null', 'bs=1', 'count=1'], errmsg="Err")
+                netperf['buffer'] = '-F '+fill_file
             except RunnerCheckError:
-                netperf['buffer'] = ''
+                if(fill_file == '/dev/urandom'):
+                    netperf['buffer'] = ''
+                else:
+                    # If the custom file is not readable, fail noisily
+                    raise RunnerCheckError("The specified test payload file does not exist or is not readable.")
 
             if not self.remote_host:
                 # only cache values if we're not executing the checks on a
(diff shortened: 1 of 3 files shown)


flent/runners.py
```diff
@@ -1022,10 +1023,14 @@ def check(self):
...
+                # If --test-payload option is specified, use data from that file
+                # else use the default value /dev/urandom. 
+                # Below sanity check will be performed in either case.
                 # Sanity check; is /dev/urandom readable? If so, use it to
                 # pre-fill netperf's buffers
-                self.run_simple(['dd', 'if=/dev/urandom', 'of=/dev/null', 'bs=1', 'count=1'], errmsg="Err")
-                netperf['buffer'] = '-F /dev/urandom'
+                fillFile = args['test_payload']
+                self.run_simple(['dd', f'if={fillFile}', 'of=/dev/null', 'bs=1', 'count=1'], errmsg="Err")
+                netperf['buffer'] = f'-F {fillFile}'
```
tohojo, line 1033: Flent runs on Python versions as old as 3.5, so we can't use f-strings, sadly...

flent/runners.py
```diff
@@ -1022,10 +1023,14 @@ def check(self):
                 netperf['-e'] = True
 
             try:
+                # If --test-payload option is specified, use data from that file
+                # else use the default value /dev/urandom. 
+                # Below sanity check will be performed in either case.
```
tohojo, line 1028: I'm not sure if the sanity check makes sense for a user-specified file: it means we'll just silently ignore the option if the file is not readable. I think it's better to fail noisily. We could do this either by doing the check but erroring out if it fails (it's the default), or we could skip the test for custom values and just let netperf fail when it can't open it.

flent/settings.py
```diff
@@ -378,6 +378,12 @@ def __call__(self, parser, namespace, values, option_string=None):
     "to decrease the socket buffer size. Can be specified multiple times, "
     "with each value corresponding to a stream of a test.")
 
+test_group.add_argument(
+    "--test-payload",
+    action="store", type=unicode, dest="TEST_PAYLOAD", default='/dev/urandom',
+    help="Path to file containing payload to pre-fill the netperf buffers with"
```
tohojo, line 384: Missing space at the end of the first string here

flent/runners.py
```diff
@@ -1022,10 +1023,14 @@ def check(self):
...
                 # Sanity check; is /dev/urandom readable? If so, use it to
                 # pre-fill netperf's buffers
-                self.run_simple(['dd', 'if=/dev/urandom', 'of=/dev/null', 'bs=1', 'count=1'], errmsg="Err")
-                netperf['buffer'] = '-F /dev/urandom'
+                fillFile = args['test_payload']
```
tohojo, line 1031: Variables are generally lowercase and underscore-separated rather then camelCase...

flent/runners.py
```diff
@@ -1022,12 +1023,19 @@ def check(self):
...
-                netperf['buffer'] = '-F /dev/urandom'
+                self.run_simple(['dd', 'if='+fill_file, 'of=/dev/null', 'bs=1', 'count=1'], errmsg="Err")
+                netperf['buffer'] = '-F '+fill_file
             except RunnerCheckError:
-                netperf['buffer'] = ''
+                if(fill_file == '/dev/urandom'):
+                    netperf['buffer'] = ''
+                else:
+                    # If the custom file is not readable, fail noisily
+                    raise RunnerCheckError("The specified fill file does not exist or is not readable.")
```
tohojo, line 1038: The option is --test-payload, so referring to "fill file" in the error message is going to confuse people. Better stick to "test payload" here as well :)

tohojo (changes requested):
That's better! One more small nit below. Also, the commit message you added is great, but please squash the two commits together so that it becomes one commit that keeps the heading of the first one, and has the text of the second commit as its body.

Also, this sentence:

> Using the option will cause netperf to fail noisily, in case the specified file is not readable or does not exist.

is slightly misleading. It's not netperf that fails noisily, it's the check in flent. So maybe "Flent will check if the file is readable before passing it to netperf, and fail the test run if it isn't" ?

PR #248, opened 2022-01-11
util: Default value of sT and Bpl for MOS score calculation

### LINKED ISSUE #247: Value of Bpl used in calculation of MOS scores for VoIP
In the mos_score function in util.py the value of Bpl used in the calculation is 1. Whereas, in the G.107 recommendation, the default recommended value is 4.3 and the permitted range is from 4.3 to 40. MOS values for some combinations of delay and loss are calculated incorrectly due to this. Setting Bpl to 4.3 will make the code conform exactly to the reference implementation given here https://www.itu.int/ITU-T/studygroups/com12/emodel/calculnb.php

Screenshot from 2022-01-10 21-06-36

hrishikeshathalye:
Changed the equations (7-27) and (7-29) by considering
the default value of sT as 1 and Bpl as 4.3 as outlined
in the G.107 recommendation. Updated the link for the
reference implementation to the latest one.

Fixes #247.

Signed-off-by: Hrishikesh Athalye <hathalye7@hotmail.com>

### DIFF
--- a/flent/util.py
+++ b/flent/util.py
@@ -367,7 +367,7 @@ def mos_score(T, loss):
     Based on ITU G.107 06/2015.
 
     Verified against the online reference implementation at
-    https://www.itu.int/ITU-T/studygroups/com12/emodelv1/calcul.php
+    https://www.itu.int/ITU-T/studygroups/com12/emodel/calculnb.php
 
     This version assumes the default values are used for all parameters other
     than delay and loss.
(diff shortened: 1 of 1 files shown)


tohojo (changes requested):
The update to Bpl is fine, but I don't think there's anything wrong with (7-27); that already considers sT as one... where are you getting the '5' from?

hrishikeshathalye:
Ah yes. I confused the 6.sT with 6-sT . Corrected.

tohojo:
Cool, that's better, but you forgot to fix the commit message :)

hrishikeshathalye:
Corrected :)

tohojo (approved):
Great, thanks for the patch!

PR #253, opened 2022-01-25
Update RRUL spec link

kasraghu:


tohojo (approved):
Thank you for the fix!

PR #262, opened 2022-02-15
gui: Don't pass floats to progressbar.setValue()

### LINKED ISSUE #261: flent-gui crashes while running test
I tried to run a rrul_be test with flent-gui, but I got the following crash:

```
[michael@localhost ~]$ flent --gui
Starting Flent 2.0.1 using Python 3.10.2.
Initialised matplotlib v3.5.1 on numpy v1.21.5.
QSocketNotifier: Can only be used with threads started with QThread
GUI loaded. Using Qt through pyqt5 v5.15.2.
qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
Starting rrul_be test. Expected run time: 310 seconds.
  File "/usr/lib/python3.10/site-packages/flent/gui.py", line 1034, in update_progress
    self.progressBar.setValue(100 * elapsed / self.total_time)
TypeError: setValue(self, int): argument 1 has unexpected type 'float'
  File "/usr/lib/python3.10/site-packages/flent/gui.py", line 1034, in update_progress
    self.progressBar.setValue(100 * elapsed / self.total_time)
TypeError: setValue(self, int): argument 1 has unexpected type 'float'
Aborted (core dumped)
[michael@localhost ~]$ Traceback (most recent call last):
  File "/usr/lib64/python3.10/multiprocessing/queues.py", line 251, in _feed
    send_bytes(obj)
  File "/usr/lib64/python3.10/multiprocessing/connection.py", line 205, in send_bytes
    self._send_bytes(m[offset:offset + size])
  File "/usr/lib64/python3.10/multiprocessing/connection.py", line 416, in _send_bytes
    self._send(header + buf)
  File "/usr/lib64/python3.10/multiprocessing/connection.py", line 373, in _send
    n = write(self._handle, buf)
```

tohojo:
The Qt ProgressBar::setValue() method expects integers and passing floats
causes a crash. Round the calculation to make sure this doesn't happen.

Fixes #261.

### DIFF
--- a/flent/gui.py
+++ b/flent/gui.py
@@ -1031,7 +1031,7 @@ class NewTestDialog(QDialog):
         if (p, s) == (0, 0):
             if not self.aborted:
                 elapsed = time.time() - self.start_time
-                self.progressBar.setValue(100 * elapsed / self.total_time)
+                self.progressBar.setValue(int(100 * elapsed / self.total_time))
         else:
             fn = os.path.join(self.settings.DATA_DIR,
                               self.settings.DATA_FILENAME)


flent/gui.py
```diff
@@ -1031,7 +1031,7 @@ def update_progress(self):
         if (p, s) == (0, 0):
             if not self.aborted:
                 elapsed = time.time() - self.start_time
-                self.progressBar.setValue(100 * elapsed / self.total_time)
+                self.progressBar.setValue(100 * elapsed // self.total_time)
```
huaracheguarache, line 1034: This still fails for me. Integer division can return a float if one or both of the variables are floats, so this should probably be changed to:

```suggestion
                self.progressBar.setValue(int(100 * elapsed // self.total_time))
```

huaracheguarache: Or would this be equivalent?

```suggestion
                self.progressBar.setValue(int(100 * elapsed / self.total_time))
```

PR #281, opened 2023-05-08
do not specify a test-specific host for the UDP_RR test

dmfreemon:
do not specify a test-specific host for the UDP_RR test


tohojo:
Please fix the commit message so it has a proper title, body and a signed-off-by tag... Otherwise LGTM :)

PR #282, opened 2023-05-08
fix for passing cong control argument to the netperf command

dmfreemon:
…f command that is executed

### DIFF
--- a/flent/runners.py
+++ b/flent/runners.py
@@ -1021,13 +1021,13 @@ class NetperfDemoRunner(ProcessRunner):
         args.setdefault('extra_test_args', "")
         args.setdefault('format', "")
         args.setdefault('marking', "")
-        args.setdefault('cong_control',
-                        self.settings.TEST_PARAMETERS.get('tcp_cong_control', ''))
         args.setdefault('socket_timeout', self.settings.SOCKET_TIMEOUT)
         args.setdefault('send_size',
                         self.settings.SEND_SIZE[0]
                         if self.settings.SEND_SIZE else "")
         args.setdefault('test_payload', self.settings.TEST_PAYLOAD)
+
+        args['cong_control'] = self.settings.TEST_PARAMETERS.get('tcp_cong_control', args.get('cong_control', ''))
 
         if self.settings.SWAP_UPDOWN:
             if self.test == 'TCP_STREAM':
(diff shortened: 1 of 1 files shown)


flent/runners.py
```diff
@@ -1021,14 +1021,14 @@ def check(self):
...
-        args.setdefault('cong_control',
-                        self.settings.TEST_PARAMETERS.get('tcp_cong_control', ''))
         args.setdefault('socket_timeout', self.settings.SOCKET_TIMEOUT)
         args.setdefault('send_size',
                         self.settings.SEND_SIZE[0]
                         if self.settings.SEND_SIZE else "")
         args.setdefault('test_payload', self.settings.TEST_PAYLOAD)
 
+        args['cong_control'] = self.settings.TEST_PARAMETERS.get('tcp_cong_control', args.get('cong_control', ''))
+
```
tohojo, line 1031: This reverses the precedence of the two sources of the arg; what are you trying to fix here?

flent/runners.py
```diff
@@ -1108,7 +1108,7 @@ def check(self):
             args['marking'] = self.parse_marking(args['marking'], "-Y {}", True)
 
         if args['cong_control']:
-            args['cong_control'] = "-K {0}".format(args['cong_control'])
+            args['cong_control'] = "-K {0},{0}".format(args['cong_control'])
```
tohojo, line 1111: Netperf already sets the same argument for both directions if there's no comma in the value, so this is not needed...

tohojo (changes requested):

flent/runners.py
dmfreemon: You are right.  I'll remove that.  Thanks for the catch.

flent/runners.py
dmfreemon: Say a test is defined like this:

     ('TCP upload',
      {'test': 'TCP_STREAM', 'length': LENGTH, 'host': HOST, 'cong_control': 'cubic',
                'units': 'Mbits/s', 'runner': 'netperf_demo'})

Note the cong_control is set in the test definition above.

Now run flent as follows:

flent test_name -H foo --test-parameter=tcp_cong_control=bbr

The value actually used will be cubic, not bbr.  I was expecting the value from the command line to override the value specified in the definition of the test.  Am I thinking about this incorrectly?  Or using the wrong approach?

tohojo: There are only a few tests that specify the congestion control algorithm, and they are supposed to override the value set on the command line. It doesn't really make sense to have the 'cubic_reno' test use bbr instead, for instance. If you want to specify the congestion control algorithm from the command line, simply use one of the tests that support this...

PR #284, opened 2023-06-30
Add Dockerfile

### LINKED ISSUE #220: Flent in Docker?
I read an intriguing article - [Run more stuff in Docker](https://jonathan.bergknoff.com/journal/run-more-stuff-in-docker/). It showed a solution to a big frustration: getting the right version of Flent, Python, netperf, and all the other libraries onto my computer without screwing up other software.

Is anyone working on a Dockerfile that bundles up Flent with its preferred support packages? This sounds like a good enhancement to the already enhanced 2.0 version. I don't have a lot of time, but might take it on in the next month to six weeks.

maybe-hello-world:
This PR adds a Dockerfile example for CLI-only usage of flent. 
The resulting docker-image is Debian (Bookworm) -based, 333MB in total.
Closes #220.

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -3,7 +3,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 RUN sed -i -e's/ main/ main contrib non-free/g' /etc/apt/sources.list.d/debian.sources \
     && apt update \
-    && apt-get install -y netperf iperf3 iputils-ping wget \
+    && apt-get install -y netperf iputils-ping irtt \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -1,0 +1,12 @@
+FROM python:slim
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN sed -i -e's/ main/ main contrib non-free/g' /etc/apt/sources.list.d/debian.sources \
+    && apt update \
+    && apt-get install -y netperf iperf3 iputils-ping wget \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN pip install --no-cache-dir matplotlib flent 
+
+ENTRYPOINT ["flent"]


packaging/docker/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:slim
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN sed -i -e's/ main/ main contrib non-free/g' /etc/apt/sources.list.d/debian.sources \
+    && apt update \
+    && apt-get install -y netperf iperf3 iputils-ping wget \
```
tohojo, line 6: Flent doesn't support iperf3, so no reason to install that. Installing irtt would be a good idea, though... Why are you installing wget?

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN pip install --no-cache-dir matplotlib flent
```
tohojo, line 10: Why not install matplotlib from the package repositories?

tohojo (changes requested):

packaging/docker/Dockerfile
maybe-hello-world: Oh, I forgot to remove wget after experiments with just binary installation trying to reduce overall size.

Does flent support iperf2? In the documentation, it's mentioned that "some tests use iperf" without specifying what particular version.

packaging/docker/Dockerfile
maybe-hello-world: Typically Debian version of matplotlib is one major version behind (the current repo version is 3.6.3 and the pip version is 3.7.1).
Do you prefer the repo version? I don't have a strong opinion here in general, up to you.

packaging/docker/Dockerfile
tohojo: Yeah, there is some limited iperf2 support; I'm not against including that in the image, but it's kinda niche so not essential either...

packaging/docker/Dockerfile
tohojo: We don't use any bleeding edge matplotlib features, so the distro package should be fine...

packaging/docker/Dockerfile
maybe-hello-world: Thanks for the clarification, I'll remove it then.

packaging/docker/Dockerfile
maybe-hello-world: Ok, I explored package dependencies for python3-matplotlib, and at the end of the day, it brings python3-dev and g++ infrastructure, which increases the image size up to 1.2GB. 
I'd prefer to stay with the PyPI version.

tohojo: Okay, yeah, that seems a bit much; okay, let's just go with the pip version then...

tohojo (approved):

maybe-hello-world:
Anything else I need to do here?

tohojo:
Roman Beltiukov ***@***.***> writes:

> Anything else I need to do here?

Nope! Let me just merge this...

dtaht:
@maybe-hello-world  * @tohojo thx for this!

PR #286, opened 2023-07-03
do not specify a test-specific host for the UDP_RR test

dmfreemon:
netperf does not support it

### DIFF
--- a/flent/runners.py
+++ b/flent/runners.py
@@ -1119,6 +1119,9 @@ class NetperfDemoRunner(ProcessRunner):
         else:
             args['socket_timeout'] = ""
 
+        if args['host']:
+            args['host'] = "-H {}".format(args['host'])
+
         if self.test in ("TCP_STREAM", "TCP_MAERTS"):
             args['format'] = "-f m"
             if args['send_size']:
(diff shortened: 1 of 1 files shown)


flent/runners.py
```diff
@@ -1138,12 +1141,13 @@ def check(self):
 
         elif self.test == 'UDP_RR':
             self.units = 'ms'
+            args['host'] = ""
```
tohojo, line 1144: This won't do the right thing if control_host is set...

dmfreemon: I don't see any tests that both use the UdpRttRunner (which sets the netperf test to UDP_RR) and also supports the --control-host option.

The netperf UDP_RR test only supports specifying the host via the global "-H" option.  netperf fails with "Invalid Argument" when the test-specific "-H" option is specified.

Can you give an example of a flent command that fails?

tohojo: The tests don't need to do anything to support it, the option is loaded directly from the settings by the runner:
```python
        args.setdefault('control_host', self.settings.CONTROL_HOST or self.host)
```

so you'll need to do a `args['control_host'] = args['host']` somewhere (and also deal with `control_local_bind` vs `local_bind`). We should probably also log a message warning if `control_host` is actually set while running a UDP_RR test, instead of just silently changing behaviour...

PR #287, opened 2023-07-04
Fixed log-scale matplotlib compatibility for v3.3

freysteinn:
The matplotlib package removed all axis directions from the base parameters for logarithmic scaling functions in v3.3.0 and up.

This commit supports both the original "basex" and "basey" parameters for older versions of matplotlib and the new "base" parameter for more recent versions of matplotlib.

### DIFF
--- a/flent/plotters.py
+++ b/flent/plotters.py
@@ -1238,7 +1238,10 @@ class Plotter(ArgParam):
             axis.set_ylim(0, top_scale)
         else:
             if self.log_base:
-                axis.set_yscale('log', basey=self.log_base)
+                try:
+                    axis.set_yscale('log', base=self.log_base)
+                except TypeError:
+                    axis.set_yscale('log', basey=self.log_base)
                 axis.set_ylim(max(0, btm_scale), top_scale)
             else:
                 axis.set_ylim(btm_scale, top_scale)
@@ -1884,7 +1887,10 @@ class CdfPlotter(Plotter):
             axis.set_xlim(left=min(min_value, axis.get_xlim()[0]))
 
         if self.log_base:
-            axis.set_xscale('log', basex=self.log_base)
+            try:
+                axis.set_xscale('log', base=self.log_base)
+            except TypeError:
+                axis.set_xscale('log', basex=self.log_base)
 
         for a, b in zip(config['axes'], self.bounds_x):
             a.set_xbound(b)


flent/plotters.py
```diff
@@ -57,6 +57,7 @@
     HAS_MATPLOTLIB = False
     MPL_VER = LooseVersion("0")
 
+
```
tohojo, line 60: Extra whitespace

flent/plotters.py
```diff
@@ -1228,7 +1229,10 @@ def _do_scaling(self, axis, data, btm, top, unit=None, allow_log=True):
             axis.set_ylim(0, top_scale)
         else:
             if self.log_base:
-                axis.set_yscale('log', basey=self.log_base)
+                if MPL_VER >= LooseVersion("3.2"):
```
tohojo, line 1232: You have an off-by-one error here; the `base` argument was introduced in matplotlib 3.3: https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.3.0.html#log-symlog-scale-base-ticks-and-nonpos-specification

Which kinda betrays that you probably didn't test this on old matplotlib versions? Please do so :)

tohojo (changes requested):

flent/plotters.py
freysteinn: Ahh, yes. I must admit that I went through the matplotlib git repo, found the commit where it changed, and made an off-by-one error. I did not test this on an old matplotlib version at all. I will do so in the future.

freysteinn:
I've updated the pull request to check for Matplotlib version 3.3 instead of the incorrect version 3.2. I also tested the commit using Python 8 and Matplotlib 3.2.2.

flent/plotters.py
```diff
@@ -1238,7 +1238,10 @@ def _do_scaling(self, axis, data, btm, top, unit=None, allow_log=True):
             axis.set_ylim(0, top_scale)
         else:
             if self.log_base:
-                axis.set_yscale('log', basey=self.log_base)
+                if MPL_VER >= LooseVersion("3.3"):
```
tohojo, line 1241: As of commit 3fb95b089917ca05b58169a055204f82039f35f2 we no longer have the LooseVersion available. I think it's better to have this check use duck typing; something like:

```python
try:
  axis.set_yscale('log', base=self.log_base)
except TypeError:
  axis.set_yscale('log', basey=self.log_base)
```

freysteinn: I've made the update and tested it again. It also worked without issues with Python 8 and Matplotlib 3.2.2.

tohojo (approved):

PR #295, opened 2024-01-17
gui: Show a message when no tabs are loaded

### LINKED ISSUE #293: Blank window on start (Flent 2.1.1, macOS)
I installed `flent` using pip on my macOS 12.7.2 machine. Running `flent-gui` from the command line was surprising:

- The `flent` window opened and displays "GUI Loaded using ..." briefly in the bar at the bottom left of the window.
- That legend disappeared.
- Nothing else happened... I wondered if there was some weird Qt/Python/system problem.
- Eventually, I farbled around in the menus and found **Run new test Cmd-R** which worked as expected.

My request: If no other action is specified, perhaps the **Run new test** could occur by default. Many thanks.

tohojo:
Instead of just showing a blank screen (which can be interpreted as the GUI
hanging), when there are no data files loaded show a message hinting that data
files can be loaded from the menu.

Fixes #293


tohojo:
Ah yes, good idea with the buttons. Pushed an update, how is this?

PR #301, opened 2024-04-03
Added support for per CPU core monitoring

freysteinn:
This commit adds multicore support to the stat_iterate.sh script and to the CpuStatsRunner. The previous script only read the first entry of the /proc/stat, which gave the average CPU usage for all cores. This modified script reads all cpu[0-9]+ records, including the average. It retains the output format of the original script; however, the implementation has some minor optimizations, which are not that significant due to how seldom the script runs.

1. It primarily relies on the awk command for the heavy lifting, making it easier to read. Awk handles reading the /proc/stat instead of the cat command.

2. The previous script used the seq command, which consumes more memory when multiple iterations occur. Awk handles this instead with a loop.

I tested the script using gawk with and without the Posix compliance flag (-P) and BusyBox v1.36.1 awk.


freysteinn:
I will have to update the unit-tests to account for the change.

freysteinn:
Do you have a recommended way of updating the test-rrul.flent.gz file to include per-CPU core records? Should I recreate the flent file? Or do you recommend doing this in another way?

tohojo:
> Do you have a recommended way of updating the test-rrul.flent.gz file to include per-CPU core records? Should I recreate the flent file? Or do you recommend doing this in another way?

test_plotters.py has a `PLOTS_MAY_FAIL` list for the plots that don't have data available in all test files. You can add the plots there; then keep the old data file around and add a new one that does contain the new data entries...

freysteinn:
> > Do you have a recommended way of updating the test-rrul.flent.gz file to include per-CPU core records? Should I recreate the flent file? Or do you recommend doing this in another way?
> 
> test_plotters.py has a `PLOTS_MAY_FAIL` list for the plots that don't have data available in all test files. You can add the plots there; then keep the old data file around and add a new one that does contain the new data entries...

I've updated the pull request by adding the per-core CPU usage entries into the `PLOTS_MAY_FAIL` variable. However, I decided not to create a new flent file because the unit tests would miss any future broken entries while the `PLOTS_MAY_FAIL` is in effect.

I think it would be better to recreate the test-rrul.flent.gz or change the PLOTS_MAY_FAIL to include which files these plots may fail in. Then, if someone accidentally breaks it, they will fail in the future.

tohojo:
Frey Alfredsson ***@***.***> writes:

> I think it would be better to recreate the test-rrul.flent.gz or
> change the PLOTS_MAY_FAIL to include which files these plots may fail
> in. Then, if someone accidentally breaks it, they will fail in the
> future.

Sounds good! Please do :)

freysteinn:
I've updated this pull request to include a list of plots to ignore per unit-test file and added a test for the new per-core monitoring functionality.

freysteinn:
I added that I tested the script with mawk 1.3.4 and added support for old versions of awk. However, I had no issues with the awk implementations I tested the script with.

tohojo (approved):
Looks good, thanks!

dtaht:
@freysteinn Wonderful thank you!

PR #307, opened 2024-06-21
Fix SyntaxWarning for '\w' '\d' in regexes

vincele:
Use raw strings literals for the affected regexes

(venv-flent) $ flent -V
/path/to/flent/flent/metadata.py:247: SyntaxWarning: invalid escape sequence '\d'
  m = re.search("(qlen|txqueuelen) (\d+)", output)
/path/to/flent/flent/metadata.py:259: SyntaxWarning: invalid escape sequence '\w'
  m = re.search("Duplex: (\w+)", output)
Starting Flent 2.1.1+git.6c8dce50 using Python 3.12.4. Flent v2.1.1+git.6c8dce50.
Running on Python 3.12.4 (main, Jun  9 2024, 22:05:49) [GCC 13.2.0]. No matplotlib found. Plots won't be available.
No usable Qt version found. GUI won't work.
ERROR: Missing test name.

### DIFF
--- a/flent/metadata.py
+++ b/flent/metadata.py
@@ -244,7 +244,7 @@ def get_link_params(iface):
         output = get_command_output("ifconfig %s" % iface)
 
     if output is not None:
-        m = re.search("(qlen|txqueuelen) (\d+)", output)
+        m = re.search(r"(qlen|txqueuelen) (\d+)", output)
         if m:
             link_params['qlen'] = m.group(2)
         m = re.search("ether ([0-9a-f:]{17})", output)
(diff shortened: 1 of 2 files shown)


vincele:
Tested with a simple 10s run:
`(venv-flent) [user@host flent]$ flent rrul -p all_scaled -l 10 -H netperf-eu.bufferbloat.net -t test -o flent.png -L logfile`

logfile contains:
2024-06-21 23:19:04,618 [flent.metadata] DEBUG: Executing 'ip link show dev enp41s0' on localhost

vincele:
I went for the minimal fix, and not touched the other regexes.

But a quick grep showed that the non-raw strings regexes are the minority, so maybe you'll want the full conversion...
I can do that, in addition to the simple fix or a single commit for the lot.

What do you think ?

tohojo:
Thanks for the fix! Yeah, we should use raw strings throughout. Please feel free to fix up all instances! :)

vincele:
also tested OK

tohojo (approved):

vincele:
Looks like the test suite failures are related to numpy and/or matplotlib.
Probably not related to the PR

PR #311, opened 2024-10-28
misc: add all progs to install target dependency

jkoeppeler:
Adding all progs as dependencies to "install" target. Otherwise the following strip command will fail.


tohojo (approved):
Looks good, thank you for the fix

PR #315, opened 2025-02-06
Update http-1up.conf with candlestick plot

eriktar:
Adding candlestick plot of bandwidth and http latency

### DIFF
--- a/flent/tests/http-1up.conf
+++ b/flent/tests/http-1up.conf
@@ -16,6 +16,19 @@ PLOTS['totals'] = {'description': 'TCP bandwidth with delay and jitter',
                    'type': 'meta',
                    'subplots': ('bandwidth', 'http_delay')}
 PLOTS['http_delay_cdf']['cutoff'] = (DELAY,-DELAY)
+
+PLOTS['box_total'] = {'description': 'Candlestick plot of HTTP latency and TCP bandwidth',
+                        'type': 'box',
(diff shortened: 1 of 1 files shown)


flent/tests/http-1up.conf
```diff
@@ -16,6 +16,19 @@ PLOTS['totals'] = {'description': 'TCP bandwidth with delay and jitter',
                    'type': 'meta',
                    'subplots': ('bandwidth', 'http_delay')}
 PLOTS['http_delay_cdf']['cutoff'] = (DELAY,-DELAY)
+
+PLOTS['box_total'] = {'description': 'Candlestick plot of HTTP latency and TCP bandwidth',
```
tohojo, line 20: We don't call them "candlestick plots" anywhere else, so let's stick with "box plot" in the description to be consistent...

flent/tests/http-1up.conf
```diff
@@ -16,6 +16,19 @@ PLOTS['totals'] = {'description': 'TCP bandwidth with delay and jitter',
...
+                        'type': 'box',
+                        'axis_labels': ['Mean bandwidth (MBits/s)', 'Mean download time (ms)'],
```
tohojo, line 22: They are not mean values when it's not a combination plot (they are samples). Let's just drop the 'axis_labels' parameter completely here.

tohojo (changes requested):
Thanks for the patch! A few comments, Please use --signoff when committing.

The same comments apply to the other PR, of course; feel free to combine them into one if that makes things easier.

PR #316, opened 2025-02-11
Adding box plots for http-1up and http-1down.

eriktar:
Keeping the axis_labels, but removing the word "mean"


tohojo (changes requested):
FYI, you can edit an existing PR by force-pushing to the branch in your own repo (`git push --force`), you don't need to create a new one. Closing the two others for now, but just for future reference.

The commit is still missing a signed-off-by tag. You can fix this with `git commit --amend --signoff`, and then force-push as described above. Please do that, but otherwise, LGTM :)

tohojo:
Merged as a16615c6a022cb573b9eaf6ffb7435b507dac981

PR #318, opened 2025-03-18
Add TCP Prague tests

dtaht:
Prague vs BBR ought to be interesting.


tohojo (changes requested):
Please write a proper commit message (including an S-o-b)...

tohojo:
There is no 'prague' CC algorithm in Linux, so this is not likely to be generally useful. Closing.

PR #321, opened 2026-05-15
fix deprecated QDockWidget::AllDockWidgetFeatures

thinrope:
QDockWidget::AllDockWidgetFeatures has been deprecated in Qt5 and missing in Qt6, it can be replaced by
DockWidgetClosable|DockWidgetMovable|DockWidgetFloatable without side-effects.

Reference:
  https://doc.qt.io/archives/qt-5.15/qdockwidget.html#DockWidgetFeature-enum
  https://doc.qt.io/qt-6/qdockwidget.html#DockWidgetFeature-enum

This may be enough to allow running flent-gui under pyqt6 :-)

$ flent-gui
Starting Flent 2.2.0 using Python 3.13.12.
Initialised matplotlib v3.10.8 on numpy v2.4.4.
GUI loaded. Using Qt through pyqt6 v6.10.3.

** NOTE ** Putting this patch directly in my [pkalin](https://github.com/thinrope/pkalin) Gentoo overlay for `net-analyser/flent-2.2.0-r1`, since `dev-python/pyqt5` has been masked.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,10 +27,11 @@ Installing Flent can be done in several ways, depending on your operating system
 
       dnf install flent
 
-- **Gentoo:**
+- **Gentoo:** (supported in `pkalin overlay <https://github.com/thinrope/pkalin>`_)
 
     .. code-block:: bash
 
+      eselect repository enable pkalin && eix-sync
       emerge net-analyzer/flent
 
 - **Nix:**
--- a/doc/quickstart.rst
+++ b/doc/quickstart.rst
@@ -19,10 +19,11 @@ Installing Flent can be done in several ways, depending on your operating system
 
       dnf install flent
 
-- **Gentoo:**
+- **Gentoo:** (supported in `pkalin overlay <https://github.com/thinrope/pkalin>`_)
 
     .. code-block:: bash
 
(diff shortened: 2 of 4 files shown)


thinrope:
Hmm, that README update was not meant to be together with this PR, since it is only loosely related, but everything is automagic these days... up to you if you want/need to split them.

tohojo (changes requested):
Very cool, thank you for the fix!

It seems to work with Qt6, but we should probably change the matplotlib backend import to `backend_qtagg` instead of the `backend_qt5agg` currently in gui.py. Care to fold that in?

I am OK with taking the README update in the same PR, but could you please add the same change to `doc/quickstart.rst`?

thinrope:
Not sure how the RPM/DEB packaging works, I guess it will need update too (I see qt5 specific packages mentioned there).

I'll try to set the proper dependency versions in my ebuild for Gentoo  now.

tohojo (approved):
Looks good, thanks!

And yeah, will have a look at the packaging scripts, and also update the CI to test with Qt6 :)

tohojo:
The failing selftests for old matplotlib versions are expected, will update the test matrix separately.

Issue #265, opened 2022-04-22
Low throughput + NetEm delay creates gaps in upload data


upnix:
**_The problem:_** In Mininet, when limiting link speed to 10Mbps (via TBF or NetEm) and adding any amount of delay with NetEm, Flent using Netperf+TCP_STREAM will return large gaps in upload data - both in CSV output and resulting charts. While Netperf acts strangely in this scenario (which I'll describe below), I believe it is Flent and the use of `apply_to` in the `DATA_SETS` data structure that causes this problem.

_**The setup:**_

- Mininet 2.3.0, installed from Github
- Flent 2.0.1 installed with `pip3 install flent`
- All on the same Ubuntu 20.0.4.4 install, directly on hardware (no VM).

With a network configuration of 1 router, 2 subnets, and 2 hosts (`h1`, `h2`), I use TBF to rate limit all links to 10Mbit/s, and NetEm to add ~28ms of delay between hosts (7ms on each link, but any amount of delay will do). I run Netserver on host `h2`, and the Flent test on `h1`, with traffic crossing the router. I'll attach my configuration files.

![image](https://user-images.githubusercontent.com/6163553/164754983-921c10b0-5c38-4f31-8586-08aef2c6e064.png)

**_Commands:_**
```
$ sudo python3 ~/mininet_networks/1Router_2Networks_3Hosts.py
mininet> h2 pkill netserver
mininet> h2 netserver
mininet> h1 ethtool -K h1-eth0 tso off gso off gro off
mininet> h2 ethtool -K h2-eth0 tso off gso off gro off
mininet> h3 ethtool -K h3-eth0 tso off gso off gro off
mininet> r0 ethtool -K r0-eth1 tso off gso off gro off
mininet> r0 ethtool -K r0-eth2 tso off gso off gro off
mininet> r0 tc qdisc add dev r0-eth1 root tbf rate 10mbit burst 4096kbit latency 5ms
mininet> r0 tc qdisc add dev r0-eth2 root tbf rate 10mbit burst 4096kbit latency 5ms
mininet> r0 tc qdisc add dev r0-eth1 parent 8001: netem delay 7ms
mininet> r0 tc qdisc add dev r0-eth2 parent 8002: netem delay 7ms
mininet> h1 tc qdisc add dev h1-eth0 root tbf rate 10mbit burst 4096kbit latency 5ms
mininet> h1 tc qdisc add dev h1-eth0 parent 8005: netem delay 7ms
[...]

PR #324, opened 2026-06-08
aggregators: Don't crash interpolating across gaps in the data

ooonea:
## Problem

The time series aggregator builds the synthetic "totals"/"avg" series by linearly interpolating each component series onto the common step grid. When a series contains a `None` value — a gap, as produced by sparse netperf output at low rates combined with latency — and that `None` ends up as an interpolation anchor, it computes `dv_dt = (v_next - v_prev) / (t_next - t_prev)` and raises:

```
TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
```

aborting the whole run in `postprocess(aggregate(...))` **before any data file is written**, so the result is lost.

This is the low-throughput-plus-latency scenario from #265 (which reports the resulting *gaps*); here a gap used as an interpolation anchor crashes the run outright. I hit it reliably running RRUL from a router through an SQM-shaped Starlink uplink (~15 Mbit/s shaped upload + ~60 ms RTT → sparse `TCP_STREAM` samples). It does not reproduce on a clean loopback run (no gaps), which is why it can go unnoticed.

## Fix

The code already guards `t_prev is None` (start/end of a series). This guards the anchor *values* too: when `v_prev` or `v_next` is `None`, leave a gap (`None`) — the same thing already done when the interpolation distance is too long — instead of crashing. The valid-data path is unchanged.

## Test

`unittests/test_aggregators.py` drives `TimeseriesAggregator.aggregate()` with a minimal series containing a `None` gap: it raises the `TypeError` on current `main` and passes with this change.

## Note

This fixes the *crash*; the gaps discussed in #265 remain (interpolating across very sparse data is not well-defined, as you noted there). Happy to adjust the approach.

### DIFF
--- a/flent/aggregators.py
+++ b/flent/aggregators.py
@@ -413,6 +413,13 @@ class TimeseriesAggregator(Aggregator):
                             result[n] = None
                         else:
                             result[n] = v_next
+                    elif v_prev is None or v_next is None:
+                        # One of the interpolation anchors is itself a gap in
+                        # the data (e.g. sparse netperf output at low rates with
+                        # latency, see issue #265). We can't interpolate across
+                        # a gap, so leave a gap in the synthetic series here too,
+                        # rather than crashing with a TypeError.
+                        result[n] = None
                     else:
                         # We found the previous and next values; interpolate
                         # between them. We assume that the rate of change dv/dt
--- a/unittests/__init__.py
+++ b/unittests/__init__.py
@@ -24,6 +24,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import os
 import unittest
 from . import test_util
+from . import test_aggregators
 from . import test_formatters
 from . import test_metadata
 from . import test_parsers
@@ -33,6 +34,7 @@ from . import test_gui
 
 
 test_suite = unittest.TestSuite([test_util.test_suite,
+                                 test_aggregators.test_suite,
                                  test_formatters.test_suite,
                                  test_metadata.test_suite,
                                  test_parsers.test_suite,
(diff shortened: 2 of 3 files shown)


tohojo (changes requested):
Thank you for the patch!

The fix looks good, but the test is not actually run as part of the test suite. Please add it to the top-level tests; something like this (on top of your patch):

```diff
diff --git i/unittests/__init__.py w/unittests/__init__.py
index 043569ccca37..5b2c97052d14 100644
--- i/unittests/__init__.py
+++ w/unittests/__init__.py
@@ -24,6 +24,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import os
 import unittest
 from . import test_util
+from . import test_aggregators
 from . import test_formatters
 from . import test_metadata
 from . import test_parsers
@@ -33,6 +34,7 @@ from . import test_gui
 
 
 test_suite = unittest.TestSuite([test_util.test_suite,
+                                 test_aggregators.test_suite,
                                  test_formatters.test_suite,
                                  test_metadata.test_suite,
                                  test_parsers.test_suite,
diff --git i/unittests/test_aggregators.py w/unittests/test_aggregators.py
index 8d21c75c4804..84cf927595b9 100644
--- i/unittests/test_aggregators.py
+++ w/unittests/test_aggregators.py
@@ -66,6 +66,7 @@ class TestTimeseriesAggregator(unittest.TestCase):
         self.assertTrue(any(v is not None for v in data),
                         "valid data points should still be interpolated")
 
+test_suite = unittest.TestLoader().loadTestsFromTestCase(TestTimeseriesAggregator)
 
 if __name__ == "__main__":
     unittest.main()
```

tohojo (approved):

